PR #5, opened 2021-01-16
[Client] Create skeleton of v0

### LINKED ISSUE #6: Create frontend v0 skeleton

jeffzh4ng:
Closes #6
Co-authored-by: Leslie Xin <leslie.xin@hotmail.com>


client/pages/_app.tsx
```diff
@@ -0,0 +1,72 @@
...
+        // events: [
+        //   {
+        //     id: 1,
+        //     name: "naruto run ring road",
+        //     description: "🏃‍♂️",
+        //   },
+        // ],
+      }}
+    >
+      {console.log("wtf is this", pageProps)}
```
jeffzh4ng, line 66: Remove this

client/pages/_app.tsx
```diff
@@ -0,0 +1,62 @@
...
+  [
+    111,
+    {
+      name: "Loo Labs",
+      description: "👩‍🔬",
+    },
+  ],
+]);
+
+function MyApp({ Component, pageProps }: AppProps) {
```
xujustinj, line 50: `MyThis` and `MyThat` naming is a pet peeve

client/pages/ClubList.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { AppContext } from "./_app";
+
+export default function ClubList() {
+  const appData = useContext(AppContext);
+  const clubListData = appData.clubs;
+
+  return (
+    <div>
+      <h1>Club List</h1>
+      {Array.from(clubListData.entries()).map(([clubId, clubInfo]) => {
```
ShreyasPrasad, line 13: Was looking into this, apparently Array.from is not supported by IE based on the newest ECMAScript standard (as detailed here): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from

However, Justin was able to load the club list in IE regardless somehow ??? Probably isn't a concern, but club list may not work on older versions of IE. Might be worth looking into.

xujustinj (commented):
image
Looks about right for a skeleton ☠️

client/pages/ClubList.tsx
```diff
@@ -0,0 +1,19 @@
...
+        return <ClubCard club={{ id: clubId, ...clubInfo }} />;
```
ShreyasPrasad, line 14: Using the spread operator and other ES6 syntax might require babel/webpack to compile into CommonJS (so it will be supported by all browsers). Not an issue for modern browsers, but important for backwards compatibility.
See: https://css-tricks.com/transpiling-es6/#:~:text=While%20support%20for%20ES6%20is,need%20to%20transpile%20our%20code.

client/pages/_app.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <AppContext.Provider
+      value={{
+        clubs
+      }}
+    >
+      <Component {...pageProps} />
+    </AppContext.Provider>
```
ShreyasPrasad, line 58: Nice use of React Contexts here!

ShreyasPrasad (changes requested):

client/pages/ClubList.tsx
xujustinj: Actually, we shouldn't be `Array.from`ing anything here. JS maps iterate in insertion order (TIL), but we don't want to use that ordering IRL. ClubListData is pretty much a cache. To control the order in which elements appear, we will need another array-like data structure of club IDs.

lesliexin: 😮 that makes sense, I guess it's just a set.

The 2 options I can think of atm are:
1. Store the clubs in an array. The con here would be linear time when searching for the club data w/ an id to display on an individual club details page.
2. Convert the map to an array similar to how we're doing here, then sort by some value (maybe alphabetical or id?). Con here would be having to sort the clubs list and finding an alt. to Array.from. 

I think option 1 is the better option imo, but lmk if there are other solutions!

client/pages/ClubList.tsx
lesliexin: (making another ticket for this to do at a later time)

client/pages/ClubList.tsx
jeffzh4ng: We will deprioritize webpack et all for another PR.

xujustinj (approved):

client/pages/ClubList.tsx
jeffzh4ng: We will stick with using a map and ensure that the backend sends a sorted list of clubs (alphabetically).

PR #8, opened 2021-01-21
🗄️ Setup Passport.js user authentication

ShreyasPrasad:
This PR consists of the implementation of essential authentication functionality, including:

-  JWT creation on account user creation (via changes to the response of the existing `create-user-use-case`)
-  JWT creation when logging in (via two new use-cases: `login-user-use-case` and `authenticate-user-use-case`)
-  token validation when accessing protected routes via the newly introduced `authenticate` method (an example exists in the v1Router along with an example protectedUserController)

Note that the tokens generated by account creation and login expire after 1 week (this can be configured from within the new `UserAuthHandler` class) and they will not succeed authentication for protected endpoints after this point (the user would be redirected to a login page/modal here). When a user logs out and logs in (while their previous token was still valid), the new token and the old token are still valid until they expire. This means that the user can have multiple active tokens; if we wish to invalidate tokens when the user logs out (and for other cases like a user resetting their password), we can look into adding a token blacklist table to our DB or Redis; this is discussed here: https://medium.com/devgorilla/how-to-log-out-when-using-jwt-a8c7823e8a6.

As a symptom of adding use cases for the above reasons, some changes to existing architecture were made:

- Adding a generic `RepoErrors` namespace for common database results such as `NotFound`, `WriteConflict` etc.
- Adding a `UserAuthHandler` class to abstract the **Passport** authentication scheme we implement away from the use cases themselves.
- Adding an auth.js class to initialize our authentication scheme (unfortunately this is a consequence of Passport)


ShreyasPrasad:
Here's a Postman file for testing (change the extension from .txt to .json after downloading. The import into P
[Waterpark.postman_collection.txt](https://github.com/loolabs/waterpark/files/6039375/Waterpark.postman_collection.txt)
ostman.

server/src/modules/users/infra/repos/implementations/mikro-user-repo.ts
```diff
@@ -12,36 +15,52 @@ export class MikroUserRepo implements UserRepo {
...
-    this.usersEntityRepo = userRepo
+  constructor(_userRepo: EntityRepository<UserEntity> | undefined) {
+    
   }
 
-  async exists(userEmail: UserEmail): Promise<boolean> {
+  async exists(userEmail: UserEmail): Promise<Result<boolean, DBErrors>> {
     const user = await DB.usersEntityRepo.findOne({ email: userEmail.value })
-    return user !== null
+    return Result.ok(user !== null)
```
jeffzh4ng, line 24: Does moving control flow outside read nicer?

`if (user) { return Result.ok(true) } else { ...}`
or
`return user !== null ? Result.ok(true) : Result.ok(false)`

server/src/modules/users/infra/repos/implementations/mikro-user-repo.ts
```diff
@@ -12,36 +15,52 @@ export class MikroUserRepo implements UserRepo {
...
   // look into async imports with JS (one option i tried exploring is inverisfy, that shit is confusing though with async modules)
-  private usersEntityRepo: EntityRepository<UserEntity> | undefined
-  constructor(userRepo: EntityRepository<UserEntity> | undefined) {
-    this.usersEntityRepo = userRepo
+  constructor(_userRepo: EntityRepository<UserEntity> | undefined) {
+    
   }
 
-  async exists(userEmail: UserEmail): Promise<boolean> {
+  async exists(userEmail: UserEmail): Promise<Result<boolean, DBErrors>> {
```
jeffzh4ng, line 22: Can `DBErrors` be more specific? Like `DBErrors.UserNotFoundError`?

server/src/modules/users/domain/value-objects/user-password.ts
```diff
@@ -64,15 +64,9 @@ export class UserPassword extends ValueObject<UserPasswordProps> {
...
-
-    if (this.isAlreadyHashed()) {
-      hashed = this.props.value
-      return this.bcryptCompare(plainTextPassword, hashed)
-    } else {
-      return this.props.value === plainTextPassword
-    }
+  public async comparePassword(hashedPassword: string): Promise<Result<boolean, UserValueObjectErrors.InvalidPasswordComparison>> {
+    if (this.isAlreadyHashed()) return Result.err(new UserValueObjectErrors.InvalidPasswordComparison("Comparing two hashed passwords"))
+    return Result.ok(await this.bcryptCompare(this.value, hashedPassword))
```
jeffzh4ng, line 69: Give the choice to the developer whether they want to read the expression or not.

```suggestion
    const passwordsMatch = await this.bcryptCompare(this.value, hashedPassword)
    return Result.ok(passwordsMatch)
```

server/src/modules/users/domain/value-objects/__tests__/userPassword.test.unit.ts
```diff
@@ -48,7 +48,7 @@ describe('UserPassword ValueObject', () => {
     expect(password.isAlreadyHashed()).toBe(false)
   })
 
-  test('When a Password is compared with its original value, it reports them to be the same', () => {
+  test('When a Password is compared with its hashed original value, it reports them to be the same', async () => {
```
jeffzh4ng, line 51:
```suggestion
  test('When a Password is compared with its original hashed value, it reports them to be the same', async () => {
```

server/src/modules/users/application/use-cases/create-user/create-user-use-case.ts
```diff
@@ -52,8 +62,21 @@ export class CreateUserUseCase implements UseCaseWithDTO<CreateUserDTO, CreateUs
...
+      
+        const tokenResponse = await new UserAuthHandler().create(updatedUser.value.id.toString())
+
+      if(tokenResponse.isErr())
+        return tokenResponse
+
+      const createUserSuccessResponse: CreateUserSuccess = {
+        user: UserMap.toDTO(user),
+        token: tokenResponse.value
+      }
```
jeffzh4ng, line 77: We need to update the use case unit test to test this new code, and all of its possible flows of control.

server/src/modules/users/application/use-cases/protected-user/protected-user-controller.ts
```diff
@@ -0,0 +1,41 @@
+import express from 'express'
+import { ControllerWithDTO } from '../../../../../shared/app/controller-with-dto'
+import { ProtectedUserUseCase } from './protected-user-use-case'
+import { ProtectedUserDTO, protectedUserDTOSchema } from './protected-user-dto'
+import { Result } from '../../../../../shared/core/result'
+import { ValidationError } from 'joi'
+
+export class ProtectedUserController extends ControllerWithDTO<ProtectedUserUseCase> {
+  constructor(useCase: ProtectedUserUseCase) { super(useCase) }
```
jeffzh4ng, line 9: The name of the use case should have some sort of verb? What's the usecase doing? `GetProtectedUserUseCase`?

jeffzh4ng (changes requested):
Looks great 🚀. I reviewed half the PR and left some feedback. I'll review the other half (all application code in  `*/application/*` in the next work session.

The main thing is we need to add unit tests for all new and updated controllers and use cases in this PR.

Flows:
- `AuthenticateUser`
- `CreateUserUse` (update it)
- `GetUser`
- `LoginUser`
- `ProtectedUser`

jeffzh4ng:
Added `do not merge` label to this PR since we will be pulling out user service into a micro service. We'll put this in a new repo, not `waterpark`.

server/src/modules/users/infra/repos/implementations/mikro-user-repo.ts
ShreyasPrasad: Yeh I agree, much more readable (was kinda impatient when coding that aha)

PR #20, opened 2021-01-28
Increase BaseController validation abstraction

xujustinj:
### Summary

Typescript wizardry for painless validation. Simply pass the schema into the constructor and it'll do the validation shenanigans automatically.

### Testing

* unit tests

### DIFF
--- a/server/src/modules/users/application/use-cases/create-user/create-user-controller.ts
+++ b/server/src/modules/users/application/use-cases/create-user/create-user-controller.ts
@@ -1,25 +1,13 @@
 import express from 'express'
-import { ValidationError } from 'joi'
 import { BaseController } from '../../../../../shared/app/base-controller'
 import { CreateUserUseCase } from './create-user-use-case'
 import { CreateUserDTO, createUserDTOSchema } from './create-user-dto'
 import { CreateUserErrors } from './create-user-errors'
 import { UserValueObjectErrors } from '../../../domain/value-objects/errors'
-import { Result } from '../../../../../shared/core/result'
 
-export class CreateUserController extends BaseController<CreateUserDTO> {
-  private useCase: CreateUserUseCase
-
+export class CreateUserController extends BaseController<CreateUserUseCase> {
   constructor(useCase: CreateUserUseCase) {
-    super()
-    this.useCase = useCase
-  }
-
-  validate(req: express.Request): Result<CreateUserDTO, ValidationError> {
-    const { error, value } = createUserDTOSchema.validate(req.body)
-
-    if (error) return Result.err(error)
-    return Result.ok(value as CreateUserDTO)
+    super(useCase, createUserDTOSchema)
   }
 
   async executeImpl(dto: CreateUserDTO, res: express.Response): Promise<express.Response> {
--- a/server/src/shared/app/base-controller.ts
+++ b/server/src/shared/app/base-controller.ts
@@ -1,8 +1,8 @@
 import * as express from 'express'
-import { BaseUseCase } from './use-case'
 import Joi, { ValidationError } from 'joi'
+import { BaseUseCase } from './base-use-case'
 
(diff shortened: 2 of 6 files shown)


server/src/shared/app/base-controller.ts
```diff
@@ -46,35 +61,35 @@ export abstract class BaseController<DTO> {
...
+  public tooMany(res: express.Response, message: string='Too many requests') {
+    return BaseController.jsonResponse(res, 429, message)
   }
 
-  public todo(res: express.Response) {
-    return BaseController.jsonResponse(res, 400, 'TODO')
+  public todo(res: express.Response, message: string='TODO') {
+    return BaseController.jsonResponse(res, 400, message)
   }
 }
```
jeffzh4ng, line 95: Beautiful

server/src/shared/app/base-controller.ts
```diff
@@ -1,23 +1,42 @@
...
+
+type extractDTO<UseCase> = UseCase extends BaseUseCase<infer T, any> ? T : never;
+
+export abstract class BaseController<UseCase extends BaseUseCase<any, any>> {
+  constructor(
+    protected readonly useCase: UseCase,
+    protected readonly schema: Joi.ObjectSchema<extractDTO<UseCase>>
+  ) {}
+
+  protected validate(body: unknown, onErr: (err: ValidationError) => void): body is extractDTO<UseCase> {
```
jeffzh4ng, line 13: Is there a way to use the `Result<T,E>` type instead of the `onErr` callback but keeping the `body is X` typeguard? We shouldn't mix error propagation styles in our codebase. We should make sure we stay consistent between

1. `throw Error()` and `try/catch`
2. using `Result<T,E>`
3. error handler with callbacks

If we can't somehow use `Result<T,E>` *and* narrow the type of `body with the type guard, then I think `onErr` or `try/catch` is ok 👍

server/src/shared/app/use-case.ts
```diff
@@ -1,3 +1,3 @@
-export interface UseCase<Request, Response> {
+export interface BaseUseCase<Request, Response> {
```
jeffzh4ng, line 1: We should rename the file as well:`base-use-case.ts`

server/src/modules/users/application/use-cases/create-user/create-user-controller.ts
```diff
@@ -1,25 +1,13 @@
...
   constructor(useCase: CreateUserUseCase) {
-    super()
-    this.useCase = useCase
-  }
-
-  validate(req: express.Request): Result<CreateUserDTO, ValidationError> {
-    const { error, value } = createUserDTOSchema.validate(req.body)
-
-    if (error) return Result.err(error)
-    return Result.ok(value as CreateUserDTO)
```
jeffzh4ng, line 22: love this

server/src/shared/app/base-controller.ts
ShreyasPrasad: default values nice

server/src/shared/app/base-controller.ts
xujustinj: https://github.com/maghis/types-joi
https://github.com/hjkcai/typesafe-joi

I don't trust either. I'm going to switch to try/catch

jeffzh4ng: i think having a callback error is better than throwing one and having to catch it with `try/catch`. At least having a callback handler is still encoded in the type system.

xujustinj: Yeah after implementing try/catch I realized it was unusable because typescript still expects something to happen if validate returns false. It can't return false, since it would just throw instead, so then we have awkward code written to handle an impossible case in the control flow

jeffzh4ng (approved):

PR #24, opened 2021-02-01
🌎 Add prettier pre-commit hook

xujustinj:
We have a `.prettierrc` file in the repo, might as well use it 🙃

Adding this pre-commit hook exposed one of our problems: we have multiple packages in a monorepo. It doesn't make sense to define multiple pre-commit hooks in multiple packages; the fix is to pull out the common dev dependencies into a top-level package.

For the most part, nothing changes aside from a pesky message when you commit (and you have to `yarn` reinstall everything after this merges).

The message looks like this:

```
PS C:\Users\xujus\CS\LooLabs\clubs> git commit -m "Test commit please ignore"
husky > pre-commit (node v12.18.3)
$ exit 0
$ yarn run lint:staged
$ lint-staged
[STARTED] Preparing...
[SUCCESS] Preparing...
[STARTED] Running tasks...
[STARTED] Running tasks for *.{js,ts,jsx,tsx}
[STARTED] prettier --write
[SUCCESS] prettier --write
[SUCCESS] Running tasks for *.{js,ts,jsx,tsx}
[SUCCESS] Running tasks...
[STARTED] Applying modifications...
[SUCCESS] Applying modifications...
[STARTED] Cleaning up...
[SUCCESS] Cleaning up...
$ exit 0
$ yarn run test:unit
$ jest --testPathPattern=unit.ts$ --maxWorkers=1
[test a2b8b20] Test commit please ignore
 1 file changed, 1 insertion(+)
```

Thanks to https://medium.com/@jsilvax/a-workflow-guide-for-lerna-with-yarn-workspaces-60f97481149d for being awesome


client/package.json
```diff
@@ -1,11 +1,18 @@
...
+  "main": "index.js",
+  "repository": "https://github.com/loolabs/waterpark",
+  "author": "Loo Labs",
+  "license": "MIT",
   "private": true,
   "scripts": {
     "dev": "next dev",
     "build": "next build",
-    "start": "next start"
+    "lint": "exit 0",
```
jeffzh4ng, line 13: Why do we have no linting for `client`?

jeffzh4ng (commented):
Looks good to me, just had one comment on why there were no linting steps declared in `client/package.json` like there is `server/package.json`. This is awesome that we have shared workspaces, this will unblock the shared API types we wanted to setup.

One thing I'll note though is that I heard Lerna gets complicated/too much power/idrk lol? One good alternative people seem to recommend is [Nx](https://nx.dev/). I think it's worth careful consideration (with a spike) before choosing the tool that facilitates our monorepo/workspace concerns.

client/package.json
xujustinj: it felt out of my jurisdiction to force prettier on the FE; that's why prettier is only installed in the BE

jeffzh4ng: I think adding prettier to the client in this PR is ok

PR #28, opened 2021-02-04
🗄 Factor the opinionated validation out of BaseController

xujustinj:
## Purpose
`BaseController` validation was being too opinionated about things:
* assuming that you have a DTO at all
* assuming that the DTO is part of the request body

This change fixes both of the above problems, as well as an outstanding bug in the `Result` implementation.

## Approach
`BaseController` is now splattered into a chain of 3 abstract classes:
* `BaseController` now makes no assumptions and its only purpose is to support HTTP requests. Endpoints that don't care about the request contents at all can extend this class.
* `TypedController<DTO> extends BaseController` specifies a DTO type; an object of that type must be built (and validated) within the abstract `buildDTO` method.
* `UseCaseController<UseCase> extends TypedController<extractDTO<UseCase>>` captures the DTO type of its corresponding `UseCase`. Most controllers will probably extend this class.

## Testing
Passed all unit tests 🧪 

#### Open Questions and Pre-Merge TODOs
- [ ] Where would user authentication would go in this chain of classes? Would it be a class in itself?
- [ ] See the Appendix for a bothersome architecture problem.

## Learning
Figuring out this structure was less about learning new techniques and more about understanding our own design.

## Appendix

`buildDTO` is meant to be implemented by calling `this.validate(...)` on things we want to type-check and then using the result. An example with a single call to `this.validate(...)` can be seen in this PR.

The intent of the `Array<ValidationError>` error type is so that we return every validation error encountered when building the DTO; this is so that we communicate requirements clearly to API users. However, this implementation requirement means we cannot short-circuit. Instead, we have to gather all the `Result`s before checking if all of them are OK. If they are, then we carry on with constructing the DTO object. If they are not, then we have to gather all of the errors.

It's not possible (at least not in a type-safe way) for us to check whether every result in an arbitrarily-sized collection is OK and subsequently use the result. Instead, we will have to write functions like `Result.resultsAllOk` to do it for arrays of fixed size. Maybe this is a reason for us to just be lazy and use short-circuiting, or maybe we should infiltrate the Typescript team and introduce smarter type-guards.

Co-authored-by: jeffzh4ng zhang.jeffreyd.@gmail.com
Co-authored-by: KTong821 <email>
Co-authored-by: ShreyasPrasad <email>

### DIFF
--- a/server/src/modules/users/application/use-cases/create-user/create-user-controller.ts
+++ b/server/src/modules/users/application/use-cases/create-user/create-user-controller.ts
@@ -1,13 +1,25 @@
 import express from 'express'
-import { BaseController } from '../../../../../shared/app/base-controller'
+import { UseCaseController } from '../../../../../shared/app/use-case-controller'
 import { CreateUserUseCase } from './create-user-use-case'
 import { CreateUserDTO, createUserDTOSchema } from './create-user-dto'
 import { CreateUserErrors } from './create-user-errors'
 import { UserValueObjectErrors } from '../../../domain/value-objects/errors'
+import { Result } from '../../../../../shared/core/result'
+import { ValidationError } from 'joi'
 
-export class CreateUserController extends BaseController<CreateUserUseCase> {
-  constructor(useCase: CreateUserUseCase) {
-    super(useCase, createUserDTOSchema)
+export class CreateUserController extends UseCaseController<CreateUserUseCase> {
+  constructor(useCase: CreateUserUseCase) { super(useCase) }
+
+  buildDTO(req: express.Request): Result<CreateUserDTO, Array<ValidationError>> {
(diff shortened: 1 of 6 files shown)


jeffzh4ng (approved):
lgtm

ShreyasPrasad (commented):
Nice abstraction with the `TypedController`. For authenticated controllers that have access to `req.user`, (or other controllers that might need access to request properties) it might make sense to either:

1. Investigate creating a base class DTO/Schema that have a user attribute and allow other DTOs to extend them so the validation that occurs in `buildDTO` will pick up on `req.user`. This depends on JOI's capabilities.
2. Add a function to `buildDTO` that validates against these additional properties in the request and does work there to change the DTO.

KTong821:
Sounds good.

PR #30, opened 2021-02-06
💻 Implement fuzzy search

lesliexin:
## Purpose
As a user, I should be able to fuzzy search through club cards by their name. 

## Approach
Implemented fuzzy search using [fuse.js](https://fusejs.io/)

ezgif com-gif-maker

### DIFF
--- a/client/pages/ClubList.tsx
+++ b/client/pages/ClubList.tsx
@@ -1,18 +1,62 @@
-import React from "react";
+import React, { useMemo, useState } from "react";
+import Fuse from 'fuse.js';
 import { useContext } from "react";
 import ClubCard from "../components";
 import { AppContext } from "./_app";
 
 export default function ClubList() {
-  const appData = useContext(AppContext);
-  const clubListData = appData.clubs;
+  const {clubs: clubListData} = useContext(AppContext);
+  const [searchValue, setSearchValue] = useState<string>("");
+
+  type Club = {
+    id: number,
+    name: string,
+    description: string
+  }
+
+  const formattedClubs: Club[] = useMemo(() => {
+    return Array.from(clubListData.entries()).map(([clubId, clubInfo]) => {
+      return ({
+        id: clubId,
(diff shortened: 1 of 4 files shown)


ShreyasPrasad (commented):
Nice use of `useMemo` for some efficiency! Clean component.

ShreyasPrasad (approved):

client/pages/ClubList.tsx
```diff
@@ -1,18 +1,62 @@
...
+  const fuse = useMemo(() => {
+      return new Fuse(formattedClubs, {
+          keys: ["name"],
+          ignoreLocation: true, // doesn't weight string location
+          threshold: 0.3 // 0=strict, 1=loose
+        })
+  }, [formattedClubs])
+
+  const clubsList = () => {
+    const filteredClubs: (Fuse.FuseResult<Club> | Club)[] = searchValue ? fuse.search(searchValue) : formattedClubs;
```
jeffzh4ng, line 36: I think our searching logic should be extracted into a hook, just to make our `ClubsList` container/page component nicer. For example, all `ClubsList` could look like this.

```
ClubsList {
  const [filteredClubs, setSearchValue] = useSearch(searchValue)

  return (
    <input onChange={(e) => setSearchValue(e.target.value)}>
    <FilteredClubsList>
  )
}
```

In my mind, these container/page components are similar to "use cases" on the backend. They are the managers that talk to the domain (hooks) and infra (react presentation components).

I created some initial ideas and guidelines we could use for client architecture here: https://www.notion.so/Architecture-63ef0af1fa7640aba6826d6664ed2642

jeffzh4ng (changes requested):

jeffzh4ng (approved):

jeffzh4ng:
Seems like we decided on keeping the filtering logic within the component. I don't feel too strongly about pulling it out into a hook as of now, but if the component grows in the future I'm pro hook :P

PR #33, opened 2021-02-06
🗄 Get all clubs endpoint

KTong821:
## Purpose
Make a working GET all clubs endpoint for V1 (I think we're calling the static version V1 now?).

## Approach
Copied Jeff's template of create users endpoint and made modifications where necessary. We've agreed that GET requests don't have DTOs so after the `BaseController` definition change all we have is a `GetAllClubsController` and no use case. 

## Testing
Wrote unit tests on `GetAllClubsController` and `Club` entity. Integration tests pending Jeff's setup. Passes all previous tests!

## Learning
Just reading through Justin's changes to `BaseController` and `CreateUserController`

## Note:

There may be parts of this PR that don't seem like they belong, such as the clubCreated event. Some of these were necessary for testing; we needed to load in from the database and "create" a club entity. Others, such as tiny changes to user tests were probably off-the-cuff remarks from Jeff to change something. Most of these were small, but let me know if there's anything you feel strongly about. Sorry about the size!


server/src/modules/clubs/domain/events/club-deleted.ts
```diff
@@ -0,0 +1,17 @@
+import { DomainEvent } from '../../../../shared/domain/events/domain-event'
```
jeffzh4ng, line 1: can we remove this file from this PR?

server/src/modules/clubs/domain/entities/club.ts
```diff
@@ -0,0 +1,57 @@
...
+import { AggregateRoot } from '../../../../shared/domain/aggregate-root'
+import { UniqueEntityID } from '../../../../shared/domain/unique-entity-id'
+import { ClubCreated } from '../events/club-created'
+import { ClubDeleted } from '../events/club-deleted'
+import { ClubId } from '../value-objects/clubId'
+
+interface ClubProps {
+  name: string
+  description: string
+  isDeleted?: boolean
```
jeffzh4ng, line 11: do we need this field for this PR?

server/src/modules/clubs/domain/value-objects/clubId.ts
```diff
@@ -0,0 +1,18 @@
+import { Result } from '../../../../shared/core/result'
+import { Entity } from '../../../../shared/domain/entity'
+import { UniqueEntityID } from '../../../../shared/domain/unique-entity-id'
+
+// ClubId IS-A Entity because we compare ClubId's not by structural properties by but id
```
jeffzh4ng, line 5: nit: `...but by id`

server/src/modules/clubs/infra/repos/club-repo.ts
```diff
@@ -0,0 +1,7 @@
+import { Club } from '../../domain/entities/club'
+
+export abstract class ClubRepo {
+  abstract exists(name: string): Promise<boolean>
+  abstract getClubByClubId(clubId: string): Promise<Club>
```
jeffzh4ng, line 5: so yeah this signature would change to include a generic mapper

server/src/modules/clubs/infra/repos/index.ts
```diff
@@ -0,0 +1,5 @@
+import { MikroClubRepo } from './implementations/mikro-club-repo'
+
+const mikroClubRepo = new MikroClubRepo() //DB.clubsEntityRepo
```
jeffzh4ng, line 3: whats this comment for?

server/src/modules/clubs/mappers/club-map.ts
```diff
@@ -0,0 +1,36 @@
+import { UniqueEntityID } from '../../../shared/domain/unique-entity-id'
+import { ClubEntity } from '../../../shared/infra/db/entities/club.entity'
+import { Club } from '../domain/entities/club'
+import { ClubDTO } from './club-dto'
+
+export class ClubMap {
+  public static toDTO(club: Club): ClubDTO { //TODO: verify use of toDTO and adapt to clubs
```
jeffzh4ng, line 7: is this TODO comment still applicable?

server/src/modules/users/domain/value-objects/user-password.ts
```diff
@@ -17,7 +17,7 @@ export class UserPassword extends ValueObject<UserPasswordProps> {
   ): Result<UserPassword, UserValueObjectErrors.InvalidPassword> {
     if (!props.hashed && !this.isAppropriateLength(props.value)) {
       return Result.err(
-        new UserValueObjectErrors.InvalidPassword(`Password doesn't meet criteria [8 chars min].`)
+        new UserValueObjectErrors.InvalidPassword(`Password doesn't meet criteria [${this.minLength} chars min].`)
```
jeffzh4ng, line 20: nice catch :)

server/src/modules/users/infra/repos/implementations/mikro-user-repo.ts
```diff
@@ -41,7 +41,7 @@ export class MikroUserRepo implements UserRepo {
   }
 
   async findAll(): Promise<Array<User>> {
-    const users = await DB.usersEntityRepo.findAll()
+    const users: Array<UserEntity> = await DB.usersEntityRepo.findAll()
```
jeffzh4ng, line 44: do we need this explicit type annotation? or can the compiler infer from the return type of `usersEntityRepo.findAll()`?

server/src/shared/infra/db/entities/club.entity.ts
```diff
@@ -0,0 +1,25 @@
...
+@Entity()
+export class ClubEntity extends BaseEntity {
+  @Property()
+  name!: string
+
+  @Property()
+  description!: string
+
+  @Property({ default: false })
+  isDeleted!: boolean
```
jeffzh4ng, line 15: do we need this column for this PR?

server/src/modules/clubs/application/use-cases/get-all-clubs/test-utils/create-club-dtos.ts
```diff
@@ -0,0 +1,13 @@
+import { ClubDTO } from '../../../../mappers/club-dto'
+
+export const createMockClubDTOs = (): Array<ClubDTO> => {
+  
+  const clubs: Array<ClubDTO> =[]
+  for (let _i = 1; _i <= 3; ++_i){
```
jeffzh4ng, line 6: nit: howcome `_i` and not `i`?

server/src/modules/clubs/application/use-cases/get-all-clubs/test-utils/create-clubs.ts
```diff
@@ -0,0 +1,19 @@
+import { Club } from '../../../../domain/entities/club'
+
+export const createMockClubs = (): Array<Club> => {
+  
+  const clubs: Array<Club> =[]
+  for (let _i = 1; _i <= 3; ++_i){
```
jeffzh4ng, line 6: same nit

server/src/modules/clubs/application/use-cases/get-all-clubs/test-utils/create-clubs.ts
```diff
@@ -0,0 +1,19 @@
...
+    const clubResult = Club.create({
+        name: `Club Name ${_i}`,
+        description: `Club Description ${_i}`
+  })
+
+  if (clubResult.isErr()) throw new Error('Something went wrong with Club creation')
```
jeffzh4ng, line 12: i think it's ok to throw here since this is being used in a test

server/src/modules/clubs/domain/entities/__tests__/club.test.unit.ts
```diff
@@ -0,0 +1,32 @@
...
+            description: "Club Description"
+        });
+    })
+    test('it adds a ClubCreated domain event on new Club creation', () => {
+        if (clubResult.isErr()) throw new Error('Club result should be isOk, not isErr')
+        expect(ClubCreated).toBeCalled()
+        expect(DomainEvents.markAggregateForDispatch).toBeCalled()
+    })
+
+    test("it adds a ClubDeleted domain event on club deletion", () => {
```
jeffzh4ng, line 26: do we need to support deleting clubs for this PR?

server/src/modules/clubs/domain/entities/club.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (isNewClub) club.addDomainEvent(new ClubCreated(club))
+
+    return Result.ok(club)
+  }
+
+  private constructor(props: ClubProps, id?: UniqueEntityID) {
+    super(props, id)
+  }
+
+  public delete(): void {
```
jeffzh4ng, line 33: do we need this for this PR?

server/src/modules/clubs/application/use-cases/get-all-clubs/get-all-clubs-controller.ts
```diff
@@ -0,0 +1,21 @@
...
+  constructor(private clubRepo: ClubRepo) {
+    super()
+  }
+
+  async execute(_req: express.Request, res: express.Response): Promise<express.Response> {
+    try {
+      const clubs = await this.clubRepo.getAllClubs()
+      const clubDTOs: Array<ClubDTO> = clubs.map((club) => ClubMap.toDTO(club))
+      return this.ok(res, clubDTOs)
+    } catch (err) {
```
jeffzh4ng, line 17: `try/catch` is what we're trying to avoid with the `Result` type. we want to encode fallibility into our type system. But right now our repos are returning promises. I also see that the initial code I put in uses a try catch too (`create-user-use-case.ts` when calling the repo.

We need to have a team discussion on how to handle fallibility from our repo layer.

1. Continue with try/catch?
2. Return `Result<T>` and modify our `Results` to have async/await support . Perhaps this is when we install a production-ready result-type package.

cc: @xujustinj

server/src/modules/clubs/application/use-cases/get-all-clubs/__tests__/get-all-clubs-controller.test.unit.ts
```diff
@@ -0,0 +1,43 @@
...
+
+describe('GetAllClubsController', () => {
+  let mockClubs: Array<Club>
+  let mockClubDTOs: Array<ClubDTO>
+
+  beforeAll(() => {
+    mockClubs = createMockClubs()
+    mockClubDTOs = createMockClubDTOs()
+  })
+
```
jeffzh4ng, line 21: we should probably add a test when the repo throws instead of resolves the correct promise, and assert that the controller returns a 500. probably wait though until we figure out what we're going to return on our repo methods as per my other comment.

jeffzh4ng (changes requested):

jeffzh4ng:
Overall looks good :) The team needs to have a discussion on what we want to return on our repo methods.

I also pointed out the removal of the notion of club deletion. I don't think we need to support this for this PR. I wonder if you might have gotten confused with us needing the `create` method though and thought we also needed a `delete`? When it comes to domain objects, `create` is a misnomer, we should really be calling it `load`, as we "load" a domain object into memory :)

jeffzh4ng:
Also, from a quick google, it seems like we do indeed want to commit lock files in our version control.

xujustinj (changes requested):
Lockfiles are supposed to be added to the repo. That surprised me too when I first heard about it, but it makes sense: the purpose of lockfiles is so that we can synchronize exactly which packages we have installed

server/src/modules/clubs/application/use-cases/get-all-clubs/get-all-clubs-controller.ts
jeffzh4ng: We wouldn't even need to wrap `.executeImpl` in a try/catch if our repos return Results instead of promises that throw. We would then need results that are `.then`able/`await`able.

server/src/modules/clubs/infra/repos/implementations/mikro-club-repo.ts
```diff
@@ -0,0 +1,35 @@
...
+  // }
+
+  async exists(name: string): Promise<boolean> {
+    const club = await DB.clubsEntityRepo.findOne({ name })
+    return club !== null
+  }
+
+  async getClubByClubId(clubId: string): Promise<Club> {
+    const club = await DB.clubsEntityRepo.findOne({ id: clubId })
+    if (!club) throw new Error() // TODO handle error
```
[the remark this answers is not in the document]
jeffzh4ng: i'm leaning towards yes, but it would be nice to have async/await support with them. i think we should look into using a production ready package now

server/src/modules/clubs/application/use-cases/get-all-clubs/__tests__/get-all-clubs-controller.test.unit.ts
xujustinj: if we don't wrap it, wouldn't an unexpected error crash the entire server?

server/src/modules/clubs/application/use-cases/get-all-clubs/get-all-clubs-controller.ts
xujustinj: if we don't wrap it, wouldn't an unexpected error crash the entire server?

server/src/modules/clubs/application/use-cases/get-all-clubs/get-all-clubs-controller.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export class GetAllClubsController extends BaseController {
+  constructor(private clubRepo: ClubRepo) {
+    super()
+  }
+
+  async execute(_req: express.Request, res: express.Response): Promise<express.Response> {
+    try {
+      const clubs = await this.clubRepo.getAllClubs()
+      const clubDTOs: Array<ClubDTO> = clubs.map((club) => ClubMap.toDTO(club))
```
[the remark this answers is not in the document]
KTong821: Yeah DTO sounds good for just before request/response.

server/src/modules/clubs/application/use-cases/get-all-clubs/test-utils/create-club-dtos.ts
KTong821: changed

server/src/modules/clubs/application/use-cases/get-all-clubs/test-utils/create-clubs.ts
KTong821: I think something similar was done in the users equivalent, so I just copied it over.

server/src/modules/clubs/domain/entities/__tests__/club.test.unit.ts
KTong821: Probably not. Delete the delete (:?

server/src/shared/app/base-controller.ts
```diff
@@ -27,35 +27,35 @@ export abstract class BaseController {
     return res.sendStatus(201)
   }
 
-  public clientError(res: express.Response, message: string='Unauthorized') {
+  public clientError(res: express.Response, message: string = 'Unauthorized') {
```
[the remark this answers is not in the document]
KTong821: I think I accidentally ran Prettier on this file.

server/src/modules/users/infra/repos/implementations/mikro-user-repo.ts
KTong821: Don't think we need it.

server/src/modules/clubs/application/use-cases/get-all-clubs/get-all-clubs-controller.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  async execute(_req: express.Request, res: express.Response): Promise<express.Response> {
+    try {
+      const result = await this.clubRepo.getAllClubs()
+      if (result.isOk()) {
+        const clubDTOs: Array<ClubDTO> = result.value.map((club) => ClubMap.toDTO(club))
+        return this.ok(res, clubDTOs)
+      } else {
+            return this.fail(res, result.error.message)
```
xujustinj, line 19:
```suggestion
        return this.fail(res, result.error.message)        
```

KTong821: not quite sure what this does..

server/src/modules/clubs/application/use-cases/get-all-clubs/get-all-clubs-controller.ts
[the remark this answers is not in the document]
ShreyasPrasad: Agree with using DTOs bidirectionally; it's something my last coop did as well. Seemed to work well.

server/src/modules/clubs/application/use-cases/get-all-clubs/get-all-clubs-controller.ts
KTong821: Final Decision:

-Remove try/catch blocks from use cases and controllers
-All repo methods contain their own try/catch blocks, and only Results are returned

server/src/modules/clubs/infra/repos/implementations/mikro-club-repo.ts
```diff
@@ -0,0 +1,23 @@
+import { QueryOrder } from '@mikro-orm/core' // {EntityRepository, QueryOrder}
```
xujustinj, line 1:
```suggestion
import { QueryOrder } from '@mikro-orm/core'
```

jeffzh4ng (approved):

xujustinj (approved):

server/src/modules/clubs/infra/repos/club-repo.ts
KTong821: final decision: leave it as is (no mapper) for now

PR #34, opened 2021-02-07
💻 Setup client side testing

jeffzh4ng:
Sets up client side testing with Jest and React Testing Library.

## Screenshots
image

## Purpose
We need testing support for our frontend.

## Approach
To test our frontend
1. `cd client`
2. `yarn test`

Followed the official Jest/React Testing Library example [0] in the nextjs repo. I also created a `test-utils.tsx` as per recommendation from [1]. This provides a custom render function that renders our app context, as well as reexports of RTL.

## Testing
_How did you test your changes?_
I added a simple test that asserts that the text "Waterpark" is shown when <Home /> is rendered.

## Learning
_Links to blog posts, patterns, libraries or addons used to solve this problem_
- [0] https://github.com/vercel/next.js/tree/master/examples/with-jest
- [1] https://testing-library.com/docs/react-testing-library/setup/

Note: I also moved our `.prettierrc` from the server into the global scope of our monorepo. My thought behind this is to keep both client and server codebases the same wrt style. But not sure if lerna has a better way of doing this.


client/config/jest/cssTransform.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
xujustinj, line 1: I've been meaning to ask about this syntax for a while; is it not possible to use our usual `export` syntax with jest?

client/tests/ClubList.test.tsx
```diff
@@ -0,0 +1,9 @@
+import Home from '../pages/index'
+import { render, screen } from '../utils/tests/test-utils'
+
+describe('Home', () => {
+  it('renders without crashing', () => {
+    render(<Home />)
+    expect(screen.getByText('Waterpark')).toBeInTheDocument()
```
xujustinj, line 7: nice example test

client/tsconfig.json
```diff
@@ -15,8 +15,8 @@
     "module": "esnext",
     "moduleResolution": "node",
     "resolveJsonModule": true,
-    "isolatedModules": true,
-    "jsx": "preserve"
+    "jsx": "preserve",
+    "isolatedModules": true
```
xujustinj, line 19: does this ordering impact any functionality? or should we just sort these properties alphabetically?

client/config/jest/cssTransform.js
jeffzh4ng: image


Just tried, and no it doesn't work. I believe this is because jest is reading that without any build step. `export` I believe is ES6 or some newer version of ecmascript that jest doesn't understand or something. The official examples also use `module.exports` syntax so I think it's ok.

client/tsconfig.json
jeffzh4ng: Changed fields of `tsconfig.json` to be in sorted order.

xujustinj (approved):

client/pages/_app.tsx
```diff
@@ -1,62 +1,62 @@
-import React from "react";
-import "../styles/globals.css";
-import type { AppProps } from "next/app";
+import React from 'react'
+import '../styles/globals.css'
+import type { AppProps } from 'next/app'
 
 interface AppData {
-  clubs: Map<ClubId, ClubInfo>;
+  clubs: Map<ClubId, ClubInfo>
```
lesliexin, line 6: we're going no semi-colons :0 should we change the frontend for consistency?

lesliexin (approved):

client/pages/_app.tsx
xujustinj: Don't worry about changing it manually; if we care enough to standardize formatting then we'll find a way to automate it!

(prettier)

PR #39, opened 2021-02-14
🗄 Restructure controllers/usecases

KTong821:
## Purpose
Restructured the controller and use-case. All controllers now have a use-case. Since not all controllers/use-cases have DTOs, we clearly split the two possibilities into WithoutDTO and WithDTO, which solves the typing and DTO validation issues for non-DTO queries such as `GET` requests. Class inheritance tree is symmetric for controllers, and there is no longer a base use-case.

## Approach
New structure completely separates DTO controllers from non-DTO controllers after earlier approaches proved too difficult to separate concerns. 

## Testing
Passes all test cases! Nothing new.


jeffzh4ng:
Can you add the file cabinet emoji to the title of this PR?

server/src/shared/app/controller-without-dto.ts
```diff
@@ -0,0 +1,10 @@
+import BaseController from './base-controller'
+import { UseCaseWithoutDTO } from './use-case-without-dto'
+
+export abstract class ControllerWithoutDTO<
```
xujustinj, line 4: It still bothers me that we're naming a class after what it isn't, rather than what it is. It's a sign that something went wrong during design, but our path to this implementation seemed pretty much inevitable?

server/src/shared/app/base-controller.ts
```diff
@@ -1,6 +1,6 @@
 import * as express from 'express'
 
-export abstract class BaseController {
+export default abstract class BaseController {
```
xujustinj, line 3:
```suggestion
export abstract class BaseController {
```

Let's stick to regular exports; [apparently default exports are bad](https://basarat.gitbook.io/typescript/main-1/defaultisbad).

server/src/shared/app/controller-with-dto.ts
```diff
@@ -0,0 +1,45 @@
+import express from 'express'
+import BaseController from './base-controller'
```
xujustinj, line 2:
```suggestion
import { BaseController } from './base-controller'
```

server/src/shared/app/controller-without-dto.ts
```diff
@@ -0,0 +1,10 @@
+import BaseController from './base-controller'
```
xujustinj, line 1:
```suggestion
import { BaseController } from './base-controller'
```

xujustinj (changes requested):

jeffzh4ng (approved):

xujustinj (approved):

PR #50, opened 2021-02-20
get all events endpoint

KTong821:
## Purpose
Added the get all events endpoint. An event currently has only "name" and "description" properties.

## Approach
Copy-pasted clubs endpoint, replaced "club" with "event". 

## Testing
Added corresponding test cases from Clubs to Events.


server/src/migrations/Migration20210220195842.ts
```diff
@@ -0,0 +1,12 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20210220195842 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('create table "event" ("id" varchar(255) not null, "created_at" timestamptz(0) not null, "updated_at" timestamptz(0) not null, "name" varchar(255) not null, "description" varchar(255) not null);');
+    this.addSql('alter table "event" add constraint "event_pkey" primary key ("id");');
+
+    this.addSql('alter table "club" drop column "is_deleted";');
```
jeffzh4ng, line 9: Why are we altering the `club` table here?

jeffzh4ng (changes requested):

server/src/migrations/Migration20210220195842.ts
KTong821: This fixes the "is_deleted" column under clubs that was accidentally added from the get-all-clubs PR. Making a new PR to clean up the clubs migration mistake.

server/src/migrations/Migration20210220195842.ts
```diff
@@ -0,0 +1,10 @@
+import { Migration } from '@mikro-orm/migrations'
```
xujustinj, line 1: this file looks like it got formatted by Prettier

xujustinj (approved):
lgtm otherwise

server/src/migrations/Migration20210220195842.ts
KTong821: fixed in latest commit.

xujustinj (approved):

jeffzh4ng (approved):

PR #57, opened 2021-02-25
💻 Adding styled-components and initial design system

### LINKED ISSUE #55: Create frontend design system
Create design system using styled-components

### LINKED ISSUE #65: Design System
Front-end: Implement design system using styled-components

lesliexin:
## Purpose
Closes #65. Adding the `styled-components` library and an initial design system.

## Approach
Creating the following files to store constants used throughout the entire application:
- `device.ts` -> various device sizes for responsivity
- `typography.ts` -> design system types
- `colours.ts` -> design system colours

#### Blog Posts
- [Comparing frontend styling options](https://www.notion.so/Frontend-Styling-9546cad80aee43619ca85c1344e6d870)


client/styles/colours.ts
```diff
@@ -0,0 +1,38 @@
+// Primary colors
```
xujustinj, line 1: If we use `colours` as the name of the object, then we should go all out 🚀 

```suggestion
// Primary colours
```

client/styles/colours.ts
```diff
@@ -0,0 +1,38 @@
...
+const purple = "#47209D";
+const blueDark = "#3370EC";
+const greenDark = "#3C8A44";
+
+// Secondary colors
```
xujustinj, line 6:
```suggestion
// Secondary colours
```

client/styles/colours.ts
```diff
@@ -0,0 +1,38 @@
...
+const blueLight = "#D8F7FE";
+const purpleLight = "#E5D0FF";
+const redLight = "#F9CED7"
+const yellowLight = "#FEF5D8";
+const greenLight = "#D8FEF7";
+
+// Neutral colors
```
xujustinj, line 13:
```suggestion
// Neutral colours
```

client/styles/device.ts
```diff
@@ -0,0 +1,22 @@
+const size = Object.freeze({
```
xujustinj, line 1: if this is specifically used for screen widths and not heights, maybe `width` is a better name?

xujustinj (changes requested):
`client/package-lock.json` shouldn't exist; try deleting the file and running `npm i` from the root folder of the repo (outside of `client`)

client/styles/colours.ts
lesliexin: oops good catch!

client/styles/device.ts
lesliexin: ya that makes sense :)

lesliexin: actually, I'm looking at it again and I feel like `size` makes more sense as we could in the future export heights if need be from the same set of constants

jeffzh4ng: Would these sizes be reusable for heights though? Or should we have two objects, `width` and `height`?

jeffzh4ng (changes requested):

client/styles/device.ts
lesliexin: hmm ya true

jeffzh4ng (approved):

client/package.json
```diff
@@ -12,12 +12,14 @@
...
   "dependencies": {
+    "@types/styled-components": "^5.1.7",
     "babel-jest": "^26.6.3",
     "fuse.js": "^6.4.6",
     "next": "10.0.3",
     "react": "17.0.1",
     "react-dom": "17.0.1",
-    "vercel-is-pull-request": "^0.0.8"
+    "vercel-is-pull-request": "^0.0.8",
+    "styled-components": "^5.2.1"
```
xujustinj, line 22: Let's keep alphabetical order

```suggestion
    "styled-components": "^5.2.1",
    "vercel-is-pull-request": "^0.0.8"
```

ShreyasPrasad (dismissed):
lgtm

xujustinj (approved):

jeffzh4ng (approved):

PR #70, opened 2021-03-13
🗄 Implement inversion of control

### LINKED ISSUE #44: Async exports
# Problem

The problem is our mikro repo singleton is being imported asynchronously. At run time, our repos are being instantiated with a null object. This sucks because the temporary fix is directly importing the singleton in our repos, which couples our repos to Mikro. We need to have support for async imports/exports.

image

This is one of the methods of our user repo where `DB` is directly imported from the DB singleton.

xujustinj:
**When reviewing this PR, pay attention to the paths of files. A lot of things moved.**

## Purpose
A sweeping refactor that closes #44, but also solves a lot of similar code smells throughout our codebase.

- burninate the global `DB` object, which was the source of our problems in #44
- convert our major application classes to allow dependency injection via the constructor (repos, use cases, controllers)
- set up similar dependency injection for helper methods that create our express routes and app
- make `Controller.execute` types a bit more precise, allowing us to get rid of an `any` cast that we have in our tests
- factored our testing utils into a `test-utils` module (test files are shorter now!)
- factored the various steps of our app construction into an `app` module (it tends to follow a very well-defined pattern of Mikro -> Mikro entity repos -> repos -> use cases -> controllers -> routers -> the app)

Not a single bit of functionality should be different.

## Approach
Uses solution 2 from [this spike](https://www.notion.so/Dependency-Injection-and-Inversion-of-Control-Containers-5117c5debb3d47ae9d47a1662e1b2f71#fd6f36dea1694211916eedd0f3e0fb71).

Not a single `index.ts` file is left alive other than our top-level monolithic one.

In the future we can use some kind of inversion of control containering, but for now `index.ts` still doesn't look _too_ bad.

## Testing
Unit and integration tests pass!

## Learning
See the spike


server/src/modules/users/infra/repos/implementations/mikro-user-repo.ts
```diff
@@ -1,47 +1,35 @@
...
+import { UserEntity } from '../../../../../shared/infra/db/entities/user.entity'
 import { UserMap } from '../../../mappers/user-map'
 import { UserRepo } from '../user-repo'
 
 export class MikroUserRepo implements UserRepo {
-  // TODO: when this class is instantiated, the entityrepo from mikroorm is undefined bc its loaded asynchronously
-
-  // right now we are directly importing the DB via module which is bad bc it couples this repo to the mikroentityrepo which is not good bc of testing. ideally we would pass this repo an entityrepostub
-
-  // look into async imports with JS (one option i tried exploring is inverisfy, that shit is confusing though with async modules)
```
jeffzh4ng, line 14: 😍

jeffzh4ng (changes requested):

server/src/modules/clubs/application/use-cases/get-all-clubs/test-utils/mock-club-dto.ts
```diff
@@ -0,0 +1,10 @@
+import { ClubDTO } from '../../../../mappers/club-dto'
+
+const createMockClubDTO = (id: any): ClubDTO => {
```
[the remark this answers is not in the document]
xujustinj: in theory we could use whatever we want as long as it can be used with string interpolation, which is pretty much `any`

server/src/__tests__/utils/environment.ts
```diff
@@ -0,0 +1,49 @@
...
+    const entities = setup.Entities()
+    this.orm = await setup.MikroORM({ debug: false })
+    this.mikroEntityRepos = setup.MikroEntityRepos(entities, this.orm)
+
+    eventEntities.forEach((eventEntity) => {
+      eventEntity.clubs.add(...clubEntities)
+      eventEntity.tags.add(...tagEntities)
+    })
+    clubEntities.forEach((clubEntity) => clubEntity.tags.add(...tagEntities))
+    await this.mikroEntityRepos.event.persist(eventEntities).flush()
```
xujustinj, line 31: This is here to match the existing `populate/teardown` implementation, but in the long run we shouldn't do this. Jest supports _global_ setup/teardown. By moving this expensive operation into the global setup, we would persist/remove the objects once per run instead of n times (where n is the number of test files).

server/src/modules/clubs/application/use-cases/get-all-clubs/test-utils/mock-club-dto.ts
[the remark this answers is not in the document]
xujustinj: nevermind, it's nicer to just restrict it to `string`

server/src/app/setup/mikro-orm.ts
```diff
@@ -1,8 +1,9 @@
-import path from 'path'
+import * as types from '../types'
```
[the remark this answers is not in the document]
xujustinj: No, I didn't even know `git mv` was a thing, the magic Git wizard 🧙 just detected it as a move.

I actually didn't even move the file, I copied over the contents of mikro-orm-config as an afterthought

server/src/modules/clubs/application/use-cases/get-all-clubs/get-all-clubs-controller.ts
```diff
@@ -7,7 +7,7 @@ export class GetAllClubsController extends ControllerWithoutDTO<GetAllClubsUseCa
     super(useCase)
   }
 
-  async execute(_req: express.Request, res: express.Response): Promise<express.Response> {
+  async execute<Res extends express.Response>(_req: express.Request, res: Res): Promise<Res> {
```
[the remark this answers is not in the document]
xujustinj: Previously, this would return a `Promise<express.Response>`, even if we passed in a more specific object. Now, the return type is the same as the type of `res`, so we can pass in more specific things (like the `httpmocks` mock response object) and it'll forward that type when it returns

server/src/app/setup/index.ts
```diff
@@ -0,0 +1,8 @@
+export { Entities } from './entities'
+export { MikroORM } from './mikro-orm'
+export { MikroEntityRepos } from './mikro-entity-repos'
+export { MikroRepos } from './mikro-repos'
```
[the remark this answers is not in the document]
xujustinj: This opens the door for us to do `/database`, `/application`, and `/http`, so I'd be down to do that kind of grouping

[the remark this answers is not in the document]
KTong821: Yeah I agree

server/src/test-utils/mocks/objects/events/mock-event-dto.ts
```diff
@@ -0,0 +1,21 @@
...
+    description: `Event Description ${id}`,
+    links: {
+      url: `Event URL ${id}`,
+      bannerImage: `Banner Image ${id}`,
+      facebook: `Facebook ${id}`,
+      twitter: `Twitter ${id}`,
+      instagram: `Instagram ${id}`,
+    },
+    startTime: '2021-01-01T00:00:00.000Z',
+    endTime: '2021-01-01T00:00:00.000Z',
```
jeffzh4ng, line 15: I wonder if the `endTime` being the same time as `startTime` will cause any problems in the future with testing?

server/src/test-utils/mocks/objects/events/mock-event.ts
```diff
@@ -0,0 +1,24 @@
+import { Event } from '../../../../modules/events/domain/entities/event'
+
+const mockEvent = (id: string): Event => {
+  const eventResult = Event.create({
+    name: `Event Name ${id}`,
+    description: `Event Description ${id}`,
+    url: `Event URL ${id}`,
+    bannerImage: `Banner Image ${id}`,
+    startTime: new Date('2021-01-01'),
+    endTime: new Date('2021-01-01'),
```
jeffzh4ng, line 10: Same question here.

server/src/test-utils/mocks/objects/clubs/mock-club-dto.ts
```diff
@@ -0,0 +1,28 @@
...
+      twitter: `Twitter ${id}`,
+      instagram: `Instagram ${id}`,
+      website: `Website ${id}`,
+    },
+    tags: ['tag1', 'tag2', 'tag3'],
+    events: [
+      {
+        name: 'Event Name 1',
+        startTime: '2021-01-01T00:00:00.000Z',
+        endTime: '2021-01-01T00:00:00.000Z',
```
jeffzh4ng, line 20: Here too?

server/src/modules/users/application/use-cases/create-user/__tests__/create-user-use-case.test.unit.ts
```diff
@@ -1,79 +1,66 @@
...
-    }
-    const createUserUseCase = new CreateUserUseCase(fakeUserRepo)
+    jest.spyOn(userRepo, 'exists').mockResolvedValue(true)
 
     const createUserResult = await createUserUseCase.execute(createUserDTO)
 
     expect(createUserResult.isErr()).toBe(true)
-    const createUserErr = createUserResult as Err<User, UserValueObjectErrors.InvalidPassword>
-    expect(createUserErr.error instanceof UserValueObjectErrors.InvalidPassword).toBe(true)
+    const createUserErr = createUserResult as Err<User, CreateUserErrors.EmailAlreadyExistsError>
```
jeffzh4ng, line 63: Nice catch

server/src/test-utils/mocks/application/mock-create-user.ts
```diff
@@ -0,0 +1,15 @@
+import { MockUserRepo } from '../../../modules/users/infra/repos/implementations/mock-user-repo'
+import { CreateUserUseCase } from '../../../modules/users/application/use-cases/create-user/create-user-use-case'
+import { CreateUserController } from '../../../modules/users/application/use-cases/create-user/create-user-controller'
+import { User } from '../../../modules/users/domain/entities/user'
+import { UserMap } from '../../../modules/users/mappers/user-map'
+
+const mockCreateUser = async (users: Array<User> = []) => {
```
jeffzh4ng, line 7:
```suggestion
const buildCreateUserMocks = async (users: Array<User> = []) => {
```

or something along those lines. `buildMocksForCreateUser` is another option?

server/src/modules/users/application/use-cases/create-user/__tests__/create-user-controller.test.unit.ts
```diff
@@ -1,122 +1,85 @@
...
-    const invalidCreateUserDTO = {
-      email: 123,
-      password: 456,
-    }
-    const mockRequest = httpMocks.createRequest({
-      body: invalidCreateUserDTO,
-    }) as DecodedExpressRequest
-    const mockResponse = httpMocks.createResponse()
-    jest.spyOn(CreateUserUseCase.prototype, 'execute').mockResolvedValue(Result.ok(mockUser))
-    const createUserController = buildController()
```
jeffzh4ng, line 68: So clean now that this is all gone. 😍

server/src/modules/users/application/use-cases/create-user/__tests__/create-user-controller.test.unit.ts
```diff
@@ -1,122 +1,85 @@
...
   })
 
   test('When the CreateUserUseCase returns Ok, the CreateUserController returns 200 OK', async () => {
-    const mockRequest = httpMocks.createRequest({
-      body: createUserDTO,
-    }) as DecodedExpressRequest
-    const mockResponse = httpMocks.createResponse()
-    jest.spyOn(CreateUserUseCase.prototype, 'execute').mockResolvedValue(Result.ok(mockUser))
-    const createUserController = buildController()
+    const user = mocks.mockUser(createUserDTO)
```
jeffzh4ng, line 25:
```suggestion
    const user = mocks.createMockUser(createUserDTO)
```

jeffzh4ng (changes requested):

server/src/test-utils/mocks/objects/events/mock-event-dto.ts
xujustinj: These new mock functions do the same thing as the old ones, just in a different file and with a different name. It's beyond the scope of this PR to actually change the tests' behaviour.

If it starts causing problems with testing in the future, the developer in the future can change it.

server/src/modules/users/application/use-cases/create-user/__tests__/create-user-controller.test.unit.ts
xujustinj: I decided on the name because I've seen people use "mock" as an informal verb in the past. `mockUser` _mocks_ a object of type `User` by returning a mock `User`. Thoughts?

jeffzh4ng: Ah, it makes sense when I read `mock` as a verb. It's just not that intuitive to me. Feel free to make the final call though.

jeffzh4ng (approved):

server/src/modules/users/application/use-cases/create-user/__tests__/create-user-controller.test.unit.ts
xujustinj: I think I'll stick with `mock` as a verb for the sake of keeping names and lines shorter

server/src/app/setup/index.ts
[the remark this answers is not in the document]
xujustinj: This made me realize that Waterpark setup actually only has three stages.

1. **db** - There are two ways to set up the DB: mikro (`setupMikroDB`) or mock (`setupMockDB`).
2. **application** - There is only one way to set up the application using our `UseCases` and `Controllers`, which is the `setupApplication` function.
3. **http** - For now, we can only choose to use Express. There are infinitely many ways we could choose the Waterpark Express API handlers. However, we only use one that features a versioned API: `setupWaterparkExpress` (we might add others in the future).

Each stage might have smaller steps within, but we can only really make meaningful decisions at the stage level. For example, it doesn't make sense to micromanage how Mikro goes from `Entity`s to `EntityRepo`s to `Repo`s. We might as well let a single factory function (`setupMikroDB`) handle all of it. What we do want to choose is whether we want a Mikro-based DB, or just a mock DB.

Just like `mockCreateUser`, the stage-level setup functions return all the intermediate objects in case the caller wants to mess with them.

docker-compose.yml
```diff
@@ -11,6 +11,8 @@ services:
       - /usr/src/app/node_modules # isolate container's node_modules from local node_modules (different OS versions)
     ports:
       - ${SERVER_PORT}:${SERVER_PORT}
+    # command: npm run test:unit
+    # command: npm run test:integration
```
jeffzh4ng, line 15: Should these lines be committed if they're commented out?

server/src/index.ts
```diff
@@ -1,40 +1,24 @@
...
-import { v1Router } from './shared/infra/http/routes'
-import mikroORMConfig from './mikro-orm.config'
-import { DB } from './shared/infra/db'
-import { UserEntity } from './shared/infra/db/entities/user.entity'
-import { ClubEntity } from './shared/infra/db/entities/club.entity'
-import { EventEntity } from './shared/infra/db/entities/event.entity'
-import { TagEntity } from './shared/infra/db/entities/tags/tag.entity'
+import { setupMikroDB } from './setup/db'
+import { setupApplication } from './setup/application'
+import { setupWaterparkExpress } from './setup/http/express'
```
jeffzh4ng, line 3: It would be clean if all three imports are imported from `./setup(/index.ts)`

server/src/index.ts
```diff
@@ -1,40 +1,24 @@
...
-  app.use('/api/v1', v1Router)
+  const { controllers } = setupApplication(repos)
 
-  app.use((_req, res) => res.status(404).json({ message: 'No route found' }))
-  app.listen(port, () => {
+  const { app } = setupWaterparkExpress(controllers, { mikroORM: orm })
+  app.listen(options.port, () => {
     console.log(`Waterpark REST API server running on http://localhost:${port}/api/v1 🦆`)
   })
 }
```
jeffzh4ng, line 20: So clean 👏

server/src/setup/http/express/basic-app.ts
```diff
@@ -0,0 +1,23 @@
+import { RequestContext } from '@mikro-orm/core'
+import express, { Router } from 'express'
+import { MikroORM } from '../../db'
+import { App } from './app'
+
+export interface BasicAppOptions {
+  mikroORM?: MikroORM
```
jeffzh4ng, line 7: Does the following code suggestion make sense? In the future we could change the type to `orm: MikroORM | TypeORM`. I guess not really since if we switch ORMs we wouldn't want to have 2 orm packages installed.

```suggestion
  orm?: MikroORM
```

server/src/setup/http/express/basic-app.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export const setupBasicApp = (expressRouter: Router, options: BasicAppOptions): App => {
+  const app = express()
+  app.use(express.json())
+
+  const entityManager = options?.mikroORM?.em
+  if (entityManager !== undefined) {
+    app.use((_req, _res, next) => RequestContext.create(entityManager, next))
+  }
```
jeffzh4ng, line 17: Can we move forwards with app setup if `entityManager === undefined` is true?

```suggestion
  } else {
    throw()?
  }
```

server/src/setup/http/express/v1-api-router.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const router = Router()
+  router.get('/', (_req, res) => {
+    return res.json({ message: 'Water, water, water. Loo, loo loo.' })
+  })
+  router.use('/clubs', endpointRouters.clubs)
+  router.use('/events', endpointRouters.events)
+  router.use('/users', endpointRouters.users)
+  return router
+}
```
jeffzh4ng, line 23: Since our API will be `v1` for the forseeable future, is extracting this file worth the cognitive load? Maybe we can move this into `./versioned-api-router.ts`?

server/src/setup/http/express/express.ts
```diff
@@ -0,0 +1,7 @@
+import { APIRouter } from './api-router'
+import { App } from './app'
+
+export interface Express {
+  apiRouter: APIRouter
+  app: App
+}
```
jeffzh4ng, line 7: Is having this type in it's own file right now worth the cognitive load? Can we move this into the file `./waterpark-express.ts`?

server/src/setup/db/db.ts
```diff
@@ -0,0 +1,5 @@
+import { Repos } from './repos'
+
+export interface DB {
```
jeffzh4ng, line 3: Can we combine this file and `./repos.ts` into one file like `./types`? Trying to figure out if splitting everything in their file is worth the cognitive load at this stage of the project.

server/src/index.ts
```diff
@@ -1,40 +1,24 @@
...
+  const migrator = orm.getMigrator()
+  await migrator.up()
 
-  app.use((_req, _res, next) => RequestContext.create(DB.orm.em, next))
-  app.use('/api/v1', v1Router)
+  const { controllers } = setupApplication(repos)
 
-  app.use((_req, res) => res.status(404).json({ message: 'No route found' }))
-  app.listen(port, () => {
+  const { app } = setupWaterparkExpress(controllers, { mikroORM: orm })
```
jeffzh4ng, line 16: The key `app` on the return type of `setupWaterparkExpress` is confusing. Since the function is always called after `setupApplication`, is isn't immediately clear what's the word "application" means in our codebase.

If you look in the definition of `setupWaterparkExpress()`, there's even a function call named `setupBasicApp`. This is confusing because I thought the "app" was setup after `setupApplication` which is from `./setup/application`.

Does renaming all instances of the word `app` in `./setup/http/express`make sense?

```suggestion
  const { expressServer } = setupWaterparkExpress(controllers, { mikroORM: orm })
```

jeffzh4ng (changes requested):

server/src/app/setup/index.ts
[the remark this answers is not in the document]
jeffzh4ng: @xujustinj Great idea! 🔥

docker-compose.yml
xujustinj: It's how I run tests. In the past I wrote and deleted those lines each time I tested the code, but this time I got tired of it and decided to just leave them commented in. My reasoning for this is that these test commands are useful for everyone, especially those who are less familiar with how to actually run the tests.

server/src/setup/http/express/basic-app.ts
xujustinj: That object has to be compatible with Mikro's `RequestContext`, so it can only be a `MikroORM`

It's an unfortunate bit of coupling where our http layer has to know about a specific persistence layer for the sake of performance

server/src/setup/http/express/basic-app.ts
xujustinj: For the sake of testing, I added mock repos in this PR. In the last commit where I added this code, there is a new `setupMockDB` function. It creates a "Mock DB" which has no ORM.

This means that it is now possible to start up Waterpark using the mock repos instead of the Mikro repos. The `RequestContext` is only needed if Mikro is used.

server/src/setup/http/express/express.ts
xujustinj: Hm, this was something I was also confused about while trying to set the files up. I put the type and the factory in the same file if there was only one sensible way to construct the type. In theory, we have infinitely many ways to create an `Express` object. The way we set up Waterpark is an arbitrary one of those ways.

As you pointed out, that seems to be leading to more confusion than it's worth. I'll put them in the same file to signify that it's the only way that _we_ will create `Express` objects

server/src/index.ts
xujustinj: Great suggestion!

jeffzh4ng (approved):

jeffzh4ng:
@xujustinj I think we're safe to merge this with my single approval. Happy to wait for another one though from Shreyas or Kevin.

server/src/modules/events/infra/repos/implementations/mikro-event-repo.ts
```diff
@@ -1,16 +1,17 @@
...
 import { AppError } from '../../../../../shared/core/app-error'
 
 export class MikroEventRepo implements EventRepo {
+  constructor(protected eventsEntityRepo: EntityRepository<EventEntity>) {}
+
   async getAllEvents(): Promise<Result<Array<Event>, AppError.UnexpectedError>> {
     try {
-      const eventEntities: Array<EventEntity> = await DB.eventsEntityRepo.find(
+      const eventEntities: Array<EventEntity> = await this.eventsEntityRepo.find(
         {},
```
ShreyasPrasad, line 15: I've seen other codebases write a single line description of the intent of a db query so its easy for devs to understand. Maybe something we want to standardize in the future for more complicated db queries, not necessary here since it's conveyed by func name.

server/src/modules/users/infra/repos/implementations/mock-user-repo.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export class MockUserRepo implements UserRepo {
+  protected userEntities: Map<string, UserEntity>
+
+  constructor(userEntities: Array<UserEntity> = []) {
+    this.userEntities = new Map(userEntities.map((userEntity) => [userEntity.id, userEntity]))
+  }
+
+  async exists(userEmail: UserEmail): Promise<boolean> {
+    console.log('exists ' + userEmail.value)
```
ShreyasPrasad, line 15: honestly, i think this is mine LOL oops. should remove though.

ShreyasPrasad (changes requested):

server/src/__tests__/utils/mikro-integration-test-environment.ts
```diff
@@ -0,0 +1,56 @@
...
+  protected clubEntities!: Array<ClubEntity>
+  protected eventEntities!: Array<EventEntity>
+  protected tagEntities!: Array<TagEntity>
+
+  public async setup(): Promise<Variables> {
+    this.mikroDB = await setupMikroDB({ debug: false })
+    const { orm, entityRepos, repos } = this.mikroDB
+
+    this.ids = [1, 2, 3].map(String)
+    this.clubEntities = this.ids.map(mocks.mockClubEntity)
```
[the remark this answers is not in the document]
xujustinj: ooh, subclasses are a great idea!

At the very least, I shouldn't assume that adding those entities is part of the _base_ integration test environment, it should be in some subclass. I'll factor that into a subclass in this PR, and that'll also serve as a signal to future devs that subclasses are encouraged 😄

server/src/modules/events/infra/repos/implementations/mikro-event-repo.ts
xujustinj: Probably out of scope for this PR, but it's a good idea for the future. Comments rarely hurt, so we should try to have more of them in general

server/src/__tests__/utils/mikro-integration-test-environment.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface Variables {
+  webServer: http.WebServer
+}
+
+// This is similar to Jest environments.
+// We don't use Jest environments here because they modify the global object in a type-unsafe way.
+class MikroIntegrationTestEnvironment {
+  protected mikroDB!: db.MikroDB
+
+  protected ids!: Array<string>
```
KTong821, line 16: Curious why these are protected if the class is never inherited from.

xujustinj: The idea is if someone were to inherit from the class, they might want to use these fields in whatever behaviour they modify

server/src/setup/http/express/types.ts
```diff
@@ -0,0 +1,10 @@
+import express from 'express'
+
+export interface APIRouter extends express.Router {}
+
+export interface WebServer extends express.Express {}
+
+export interface Express {
```
KTong821, line 7: Pretty nit, but a more descriptive name would've made understanding the interactions easier. Up to you.

KTong821 (dismissed):
Very clean, great job. Pretty easy to understand where things fall into place after hunting down the imports (:  👍👍

server/src/__tests__/utils/mikro-integration-test-environment.ts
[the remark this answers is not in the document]
KTong821: Every time Justin kills a class 5 more appear. 

Not sure if we should include in this PR. If it's definitely part of IoC then go ahead, but I think we can have further breakdown of the testing environment in a separate PR. Just my two cents.

server/src/setup/http/express/types.ts
xujustinj: I also dislike the name! The main problem is that anything with "App" in it conflicts with `application`, otherwise I'd totally call it `ExpressApp`.

How about `WaterparkExpress`?

server/src/__tests__/utils/mikro-integration-test-environment.ts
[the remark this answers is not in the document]
xujustinj: I think it is in the scope of this PR because this is actually the PR that introduced the testing environment in the first place. (The old `DB` object doesn't count as an environment.) I might as well get it right the first time.

server/src/test-utils/environments/mock-entity-mikro-environment.ts
```diff
@@ -0,0 +1,44 @@
+import { mocks } from '..'
+import { ClubEntity } from '../../shared/infra/db/entities/club.entity'
+import { EventEntity } from '../../shared/infra/db/entities/event.entity'
+import { TagEntity } from '../../shared/infra/db/entities/tags/tag.entity'
+import { MikroEnvironmentVariables, MikroTestEnvironment } from './mikro-environment'
+
+export class MockEntityMikroTestEnvironment extends MikroTestEnvironment {
```
jeffzh4ng, line 7: What other test environment subclasses could we potentially have in the future?

jeffzh4ng (commented):
Looks great, LGTM!

jeffzh4ng (approved):

ShreyasPrasad (approved):

server/src/test-utils/environments/mock-entity-mikro-environment.ts
xujustinj: One example is if we want to set up only users, then all the events and clubs don't make much sense.

Issue #46, opened 2021-02-18
Create endpoint for Clubs


jeffzh4ng:
# Motivation

`GET /clubs`

PR #71, opened 2021-03-13
🗄 V0 Clubs and Events Endpoints

### LINKED ISSUE #45: Create endpoint for Events
# Motivation

`GET /events/`

KTong821:
## Purpose
Resolve #45 and #46. Fill out the clubs and events endpoint according to the API defined on the Figma. Apply migration to DB and update unit tests to match. 

## Testing
Same unit tests as before, just adding new fields to the test_utils functions.

Update: Added new integration tests for clubs and events. Waiting for integration testing to be set up, so environment variable DATABASE_URL has to be manually set for now.

## Learning

MikroORM relationships:

Many-to-many can be bidirectional: owning side is denoted using "inversedBy" and owned side is denoted using "mappedBy". Relations are lazy loaded by default (ie. orm.find(clubs, ...) returns a club. Properties like club.name are immediately defined, but a property defined as a relationship like clubs.tags would be undefined). To prevent multiple db calls (done using club.tags.init() for example), the strategy "LoadStrategy.JOINED" was used, to auto-load in a single db call. There's an easy option to have this be the case for all of MikroORM, but that's something to decide on later. 

## Notes

- I think this PR should set the standard of using options objects for all MikroORM relationships.
- Right now, clubs and events each store skinny versions of events and clubs respectively (typed as BasicEvent and BasicClub). In terms of domain, there isn't a clear divide right now because as mentioned earlier, the ORM just directly pulls up whatever's needed, instead of the use-cases talking to each other. So a TODO for V1.
- Because Jest is picky about comparing strings to dates, all DTOs that have Date objects in them get coverted to strings on our side using Date.toJSON() instead of the json function call inside the controller. A bit messy, but not too significant I think.


server/src/shared/infra/db/entities/club.entity.ts
```diff
@@ -11,6 +13,33 @@ export class ClubEntity extends BaseEntity {
...
   description!: string
 
+  @Property()
+  size!: number
+
+  @Property({ nullable: true })
+  backgroundImageURL!: string | null
+
+  @Property({ nullable: true })
+  iconURL!: string | null
```
jeffzh4ng, line 23: What's the difference between decorating the property with `@Property({ nullable: true })` and then marking the type `iconURL!: string | null`?

server/src/modules/clubs/domain/entities/club.ts
```diff
@@ -7,6 +7,22 @@ import { ClubId } from '../value-objects/clubId'
 interface ClubProps {
   name: string
   description: string
+  backgroundImageURL: string | null
+  iconURL: string | null
```
xujustinj, line 11: super nit: these are not in the same order as they appear in the `Club` class

server/src/shared/infra/db/entities/tags/tag.entity.ts
```diff
@@ -0,0 +1,15 @@
+import { Collection, Entity, ManyToMany, Property } from '@mikro-orm/core'
+import { BaseEntity } from '../base.entity'
+import { ClubEntity } from '../club.entity'
+import { EventEntity } from '../event.entity'
+
+@Entity()
+export class TagEntity extends BaseEntity {
+  @Property()
+  name!: string
+  @ManyToMany({ entity: () => ClubEntity, inversedBy: "tags" })
```
jeffzh4ng, line 10:
```suggestion

  @ManyToMany({ entity: () => ClubEntity, inversedBy: "tags" })
```

server/src/modules/clubs/mappers/club-dto.ts
```diff
@@ -1,4 +1,21 @@
+
```
xujustinj, line 1:
```suggestion
```

server/src/modules/clubs/mappers/club-map.ts
```diff
@@ -1,21 +1,58 @@
 import { UniqueEntityID } from '../../../shared/domain/unique-entity-id'
 import { ClubEntity } from '../../../shared/infra/db/entities/club.entity'
-import { Club } from '../domain/entities/club'
-import { ClubDTO } from './club-dto'
+import { EventEntity } from '../../../shared/infra/db/entities/event.entity'
+import { BasicEvent, Club } from '../domain/entities/club'
+import { ClubDTO, BasicEventDTO } from './club-dto'
+
+let basicEventMap = async (events: Array<EventEntity>) => {
```
jeffzh4ng, line 7:
```suggestion
const basicEventMap = async (events: Array<EventEntity>): Promise<Array<BasicEvent>> => {
```

xujustinj, line 7: Can this be const?

```suggestion
const basicEventMap = async (events: Array<EventEntity>) => {
```

Also I'm guessing we don't want this method to be in a fully-fledged Map class because `BasicEvent` isn't technically a domain object?

server/src/modules/events/domain/entities/__tests__/event.test.unit.ts
```diff
@@ -13,8 +13,19 @@ describe('Event AggregateRoot', () => {
...
+      url: 'Event URL',
+      startTime: new Date('2021-01-01'),
+      endTime: new Date('2021-01-01'),
+      facebookLink: 'Facebook',
+      twitterLink: 'Twitter',
+      instagramLink: 'Instagram',
+      websiteLink: 'Website',
+      backgroundImageURL: 'Background URL',
+      tags: ['tag1', 'tag2', 'tag3'],
+      clubs: [{ name: 'Club Name 1', iconURL: '' }],
```
ShreyasPrasad, line 25: This seems like code we might repeat in the future when generating events; that raises the question, for larger objects like an Event, should we formalize the creation of factory (mock) methods so we can avoid convoluting tests with large creation calls? Not sure if this is actually necessary, since we have create-event-dtos already. This just might be a consequence of the unit test.

server/src/modules/events/mappers/event-dto.ts
```diff
@@ -1,4 +1,16 @@
+import { BasicClub } from "../domain/entities/event";
+
 export interface EventDTO {
```
jeffzh4ng, line 3: If `event-map.ts` is the only consumer of `EventDTO` I think we should just move this type into `event-map.ts`.

jeffzh4ng (changes requested):

server/src/modules/clubs/infra/repos/implementations/mikro-club-repo.ts
```diff
@@ -14,7 +14,11 @@ export class MikroClubRepo implements ClubRepo {
         {},
         { orderBy: { name: QueryOrder.DESC_NULLS_LAST } }
       )
-      const clubs = clubEntities.map((clubEntity: ClubEntity): Club => ClubMap.toDomain(clubEntity))
+      const clubs = await Promise.all(
+        clubEntities.map(
+          async (clubEntity: ClubEntity): Promise<Club> => await ClubMap.toDomain(clubEntity)
+        )
+      )
```
xujustinj, line 21:
```suggestion
      const clubs = await Promise.all(clubEntities.map(ClubMap.toDomain))
```

xujustinj (changes requested):
rest lgtm

jeffzh4ng:
Can you change the PR title (and later, commit message) to be more descriptive?

server/src/modules/clubs/mappers/club-map.ts
KTong821: I could make a separate file for it, but this Map class is the only thing that uses it, so I thought might as well just have it at the top. Made const.

server/src/__tests__/event.test.integration.ts
```diff
@@ -0,0 +1,27 @@
...
+app.use('/api/v1', v1Router)
+
+describe('Event Router', () => {
+  beforeAll(async () => {
+    await populate()
+  })
+
+  afterAll(async () => {
+    await teardown()
+  })
```
xujustinj, line 17:
```suggestion
  beforeAll(populate)
  afterAll(teardown)
```
I think this works?

xujustinj: likewise for all of the other integration test files

server/src/__tests__/user.test.integration.ts
```diff
@@ -20,8 +24,8 @@ describe('User Router', () => {
     DB.usersEntityRepo = DB.orm.em.getRepository(UserEntity)
   })
 
-  afterAll(() => {
-    DB.orm.close()
+  afterAll(async () => {
+    await DB.orm.close()
   })
```
xujustinj, line 29:
```suggestion
  afterAll(DB.orm.close)
```

server/src/__tests__/utils/create-tag-entities.ts
```diff
@@ -0,0 +1,12 @@
+import { TagEntity } from '../../shared/infra/db/entities/tags/tag.entity'
+
+export const createMockTagEntities = (): TagEntity[] => {
+  const tags: Array<TagEntity> = []
+  for (let i = 1; i <= 3; ++i) {
+    const tagEntity = new TagEntity()
+    tagEntity.name = `Tag Name ${i}`
+
+    tags.push(tagEntity)
+  }
```
xujustinj, line 10: we should probably instead have a `mockTagEntity` factory function and then return `[1,2,3].map(id => mockTagEntity(id))`, but we can defer this and do it in #70 instead

server/src/modules/clubs/mappers/club-map.ts
```diff
@@ -1,21 +1,78 @@
...
+
+interface BasicEventDTO {
+  name: string
+  startTime: string
+  endTime: string
+  backgroundImageURL?: string
+  tags: Array<string>
+}
+
+export interface ClubDTO {
```
xujustinj, line 14: I think it was fine to keep this interface in the `club-dto.ts` file, it feels out-of-scope in a file about data transformations

server/src/modules/clubs/mappers/club-map.ts
```diff
@@ -1,21 +1,78 @@
...
+      twitterLink: club.twitterLink,
+      instagramLink: club.instagramLink,
+      websiteLink: club.websiteLink,
+      tags: club.tags,
+      events: club.events.map((event) => {
+        return {
+          ...event,
+          startTime: event.startTime.toJSON(),
+          endTime: event.endTime.toJSON(),
+        } as BasicEventDTO
```
xujustinj, line 58: Is there any type-safe way to do this?

server/src/modules/clubs/mappers/club-map.ts
```diff
@@ -1,21 +1,78 @@
...
+      }),
     }
   }
 
-  public static toDomain(clubEntity: ClubEntity): Club {
+  public static async toDomain(clubEntity: ClubEntity): Promise<Club> {
```
xujustinj, line 63: random thought - every time we make a function async, we have to change a bunch of its usages across the codebase... I can't think of any way around this apart from making everything async by default

server/src/modules/clubs/mappers/club-map.ts
```diff
@@ -1,21 +1,78 @@
...
+const basicEventMap = async (events: Array<EventEntity>): Promise<Array<BasicEvent>> => {
+  const eventPromises = events.map(async (event) => {
+    await event.tags.init()
+    return {
+      name: event.name,
+      startTime: event.startTime,
+      endTime: event.endTime,
+      backgroundImageURL: event.backgroundImageURL,
+      tags: event.tags.getItems().map((tag) => tag.name),
+    } as BasicEvent
```
xujustinj, line 36: Is there any type-safe way to do this?

server/src/modules/events/mappers/event-map.ts
```diff
@@ -16,6 +41,18 @@ export class EventMap {
...
+        startTime: eventEntity.startTime,
+        endTime: eventEntity.endTime,
+        facebookLink: eventEntity.facebookLink,
+        twitterLink: eventEntity.twitterLink,
+        instagramLink: eventEntity.instagramLink,
+        websiteLink: eventEntity.websiteLink,
+        backgroundImageURL: eventEntity.backgroundImageURL,
+        tags: eventEntity.tags.getItems().map((tag) => tag.name),
+        clubs: eventEntity.clubs.getItems().map((club) => {
+          return { name: club.name, iconURL: club.iconURL } as BasicClub
```
xujustinj, line 54: Is there any type-safe way to do this?

xujustinj (changes requested):

server/package.json
```diff
@@ -12,7 +12,7 @@
     "start:dev": "concurrently \"tsc -w\" \"nodemon dist/index.js\"",
     "test:all": "jest --maxWorkers =1",
     "test:unit": "jest --testPathPattern=unit.ts$ --maxWorkers=1",
-    "test:integration": "jest --testPathPattern=integration.ts$ --maxWorkers=1"
+    "test:integration": "jest --testPathPattern=integration.ts$ --runInBand"
```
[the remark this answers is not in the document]
KTong821: Oh I was messing around with that trying to fix an async error. Changed back after tests pass on maxWorkers=1

server/src/__tests__/user.test.integration.ts
KTong821: DB.orm.close doesn't work as a callback. The error if you're curious:

Argument of type '(force?: boolean | undefined) => Promise<void>' is not assignable to parameter of type 'ProvidesCallback'.
  Types of parameters 'force' and 'cb' are incompatible.
    Type 'DoneCallback' is not assignable to type 'boolean | undefined'.
      Type 'DoneCallback' is not assignable to type 'true'.

server/src/__tests__/club.test.integration.ts
```diff
@@ -0,0 +1,27 @@
...
+app.use('/api/v1', v1Router)
+
+describe('Club Router', () => {
+  beforeAll(async () => {
+    await populate()
+  })
+
+  afterAll(async () => {
+    await teardown()
+  })
```
jeffzh4ng, line 17: clean af 👌

server/src/shared/infra/db/entities/club.entity.ts
KTong821: Removed "| null" in latest, it was to fix a type error that I later changed.

server/src/modules/clubs/mappers/club-map.ts
jeffzh4ng: yeah why are these methods `async`?

server/package.json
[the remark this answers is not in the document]
KTong821: Scratch that, both are now --runInBand

server/src/modules/clubs/mappers/club-map.ts
KTong821: Okay, I reverted to having DTO files for both clubs and events.

server/src/modules/clubs/mappers/club-map.ts
KTong821: I changed the definition for Basic Event to have the optional fields match that of the Entity definition, and that let me remove the type cast. A TODO we discussed was on having nullable types rather than undefined types, but MikroORM isn't happy about that.

@Property({ nullable: true })
  facebookLink: string

Property 'facebookLink' has no initializer and is not definitely assigned in the constructor.

A cursed way:

@Property({ nullable: true })
  facebookLink!: string | null

The compiler doesn't get mad, but this is pretty confusing.

server/src/modules/events/mappers/event-map.ts
KTong821: Removed.

server/src/modules/clubs/domain/entities/club.ts
KTong821: fixed.

server/src/modules/clubs/mappers/club-map.ts
xujustinj: in this case there's an async call in the function body to `basicEventMap`

server/src/modules/events/mappers/event-dto.ts
xujustinj: after further discussion we decided against this

xujustinj (approved):

xujustinj (changes requested):
actually nvm, I'm seeing a few outstanding change suggestions that need to be resolved before this is approved

xujustinj (changes requested):

server/src/shared/infra/db/entities/tags/tag.entity.ts
xujustinj: @KTong821 can you accept this suggestion?

server/src/modules/clubs/application/use-cases/get-all-clubs/__tests__/get-all-clubs-use-case.test.unit.ts
```diff
@@ -3,7 +3,7 @@ import { MikroClubRepo } from '../../../../infra/repos/implementations/mikro-clu
 import { GetAllClubsUseCase } from '../get-all-clubs-use-case'
 import { createMockClubs } from '../test-utils/create-clubs'
 import { createMockClubDTOs } from '../test-utils/create-club-dtos'
-import { ClubDTO } from '../../../../mappers/club-dto'
+import { ClubDTO } from '../../../../mappers/club-map'
```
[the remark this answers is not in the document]
KTong821: Fixed.

server/src/shared/domain/aggregate-root.ts
```diff
@@ -38,4 +42,4 @@ export abstract class AggregateRoot<T> extends Entity<T> {
   get domainEvents(): DomainEvent[] {
     return this._domainEvents
   }
-}
+}
```
xujustinj, line 45:
```suggestion
}

```

.gitignore
```diff
@@ -6,7 +6,7 @@
 
 # Junk
 .DS_Store
-
+.vscode
```
xujustinj, line 10:
```suggestion


```

xujustinj (approved):

server/src/modules/clubs/domain/entities/club.ts
```diff
@@ -7,6 +7,22 @@ import { ClubId } from '../value-objects/clubId'
 interface ClubProps {
   name: string
   description: string
+  iconImage: string
+  bannerImage: string
+  facebook?: string
+  twitter?: string
+  instagram?: string
+  website?: string
```
jeffzh4ng, line 15: I feel like the domain type should change too? All these fields being under `links`?

jeffzh4ng (changes requested):

server/src/modules/clubs/domain/entities/club.ts
KTong821: changed in latest pr

jeffzh4ng (dismissed):

xujustinj (dismissed):

xujustinj (approved):
LGTM!

jeffzh4ng (approved):

PR #73, opened 2021-03-18
💻 Upgrade packages, specifically Next.js

### LINKED ISSUE #52: Upgrade Next version once issue 20456 is resolved
See [this comment](https://github.com/loolabs/waterpark/pull/38#issuecomment-777257494) for context.

**Next** issue can be tracked [here](https://github.com/vercel/next.js/issues/20456)

xujustinj:
Recently noticed that new `next` versions have been released, fixing the `sharp` dependency problem

Kind of infuriating that we have to upgrade everything else... so much for lock files

Closes #52

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -15,7 +15,7 @@
     "@types/styled-components": "^5.1.7",
     "babel-jest": "^26.6.3",
     "fuse.js": "^6.4.6",
-    "next": "10.0.3",
+    "next": "^10.0.9",
     "react": "17.0.1",
     "react-dom": "17.0.1",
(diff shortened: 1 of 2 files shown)


jeffzh4ng (approved):

jeffzh4ng:
I think the emoji should be client-side, not global?

lesliexin (approved):

xujustinj:
> I think the emoji should be client-side, not global?

It also affects the packages installed for `server`, so it's technically global. Then again, the point of the update was to just upgrade Next.js, so I'll update the icon to client just to match our intent.

PR #74, opened 2021-03-18
💻 Generalize types and utilities

xujustinj:
## Screenshots
No visual or functional changes

## Purpose

Random changes to streamline stuff in preparation for the FE blitz.

- Fix `npm run test` on `client`
- Generalize some helper functions and hooks
- Add `id` to `Club` and `Event` types
- Run `Prettier` on the client files affected by any of the above changes

## Testing
Verified UI looks the same (searching still works)

Our single frontend test passes

#### Open Questions and Pre-Merge TODOs
- Not sure if there's a use case for having a `Club` or `Event` without an `id`, which would motivate us keeping `Id` separate. Maybe when the user is creating one in the UI? I think creation would involve a completely different DTO though, so this would be fine?

#### Blog Posts


xujustinj:
Apology in advance for the hard-to-review code. Prettier really destroyed some of these files with cosmetic changes...

client/components/clubs/ClubList.tsx
```diff
@@ -1,39 +1,26 @@
...
 interface ClubListProps {
-  clubListData: Map<Id, ClubInfo>;
+  clubs: Map<Id, Club>
 }
 
-const clubsList = (filteredClubs: SearchResult) => {
-  return filteredClubs.map((club: SearchItem | { item: SearchItem }) => {
-    if ("item" in club) {
-      club = club.item as SearchItem;
-    }
```
jeffzh4ng, line 13: What happened to this logic?

jeffzh4ng (changes requested):

client/context/Base.tsx
```diff
@@ -1,83 +1,85 @@
...
+  description: string
 }
 
-export interface EventInfo {
-    name: string
-    description: string
+export type Event = {
+  id: Id
+  name: string
+  description: string
```
lesliexin, line 20: omg adding the ID to the Club and Event types simplified so much 🙌, I can't remember why we separated it in the first place..

jeffzh4ng:
@xujustinj We're going to merge this in today because it's blocking the rest of us. We can discuss it later (yolo)

jeffzh4ng (approved):

lesliexin (approved):

client/components/clubs/ClubList.tsx
xujustinj: It's because the changes to searchResult standardized its return type to always be an array of the same type as what was passed in:

```
const searchResult: Array<T> = searchValue
  ? fuse.search(searchValue).map(({ item }: Fuse.FuseResult<T>) => item)
  : items
```

Now, when we take the results from the search hook, they are always `Club[]` rather than `(FuseThingy | Club)[]`, so there's no need to adjust based on the type of the array element

client/utils/IndexUtils.tsx
```diff
@@ -0,0 +1,3 @@
+export const indexData = <K, V extends { id: K }>(items: Array<V> = []): Map<K, V> => {
+  return new Map<K, V>(items.map((item) => [item.id, item]))
+}
```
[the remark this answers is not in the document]
xujustinj: Good point! I figured we'd only be using this for generating mock data so we can just hard-code everything.

We could set up defaults with [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign), but things would no longer be type-safe.

PR #76, opened 2021-03-20
💻 Fixing mobile logo showing on initial load

lesliexin:
## Screenshots
image

## Purpose
Initially was loading the mobile logo. https://recordit.co/oVL7wKg5nY.

## Approach
Dynamically importing the navbar is the only way to ensure the correct logo is displayed on the initial load.
https://nextjs.org/docs/advanced-features/dynamic-import#with-no-ssr

### DIFF
--- a/client/components/hooks/UseViewport.tsx
+++ b/client/components/hooks/UseViewport.tsx
@@ -1,13 +1,13 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useState } from 'react'
 
 export const useViewport = () => {
-    const [width, setWidth] = useState(0);
+  const [width, setWidth] = useState(window.innerWidth)
 
-    useEffect(() => {
-        const handleWindowResize = () => setWidth(window.innerWidth);
-        window.addEventListener("resize", handleWindowResize);
-        return () => window.removeEventListener("resize", handleWindowResize);
-    }, []);
+  useEffect(() => {
+    const handleWindowResize = () => setWidth(window.innerWidth)
+    window.addEventListener('resize', handleWindowResize)
+    return () => window.removeEventListener('resize', handleWindowResize)
+  }, [])
 
-    return { width };
+  return { width }
 }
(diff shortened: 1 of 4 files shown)


client/components/hooks/UseViewport.tsx
```diff
@@ -1,13 +1,13 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useState } from 'react'
 
 export const useViewport = () => {
-    const [width, setWidth] = useState(0);
```
jeffzh4ng, line 4: ah that bug makes sense now

lesliexin: looool ya

client/pages/_app.tsx
```diff
@@ -1,13 +1,17 @@
 import React from 'react'
 import '../styles/globals.css'
 import type { AppProps } from 'next/app'
-import { AppProvider } from "../context"
-import { NavBar } from "../components/NavBar"
+import { AppProvider } from '../context'
+import dynamic from 'next/dynamic'
+
+const DynamicNavBar = dynamic(() => import('../components/NavBar').then((mod) => mod.NavBar), {
```
jeffzh4ng, line 7: What are dynamic imports?

jeffzh4ng (approved):

client/pages/_app.tsx
lesliexin: When we specify `ssr: false` it let's us load the component on the client-side, it's the only way to get the right initial window size on initial load unfortunately

ShreyasPrasad: Never seen dynamic used before either; always just handled navbar elements with media queries. This is a cool approach though.

ShreyasPrasad (approved):

jeffzh4ng (approved):

ShreyasPrasad (approved):

PR #80, opened 2021-04-01
💻 Create Club Details page

### LINKED ISSUE #62: Club Detail Page
As a Waterloo student, I want to dig deeper into a particular club so I can learn specific information about them

jeffzh4ng:
## Screenshots
image

Video: https://recordit.co/vS89Jq2wCc

## Purpose
Closes #62

## Testing
Since this is a static page with no user reaction, no tests were written.


client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+import { faLongArrowAltLeft } from '@fortawesome/free-solid-svg-icons'
+import { BasicEvent } from '../../context/Base'
+import moment from 'moment'
+
+interface BannerProps {
+  backgroundImageUrl: string
+}
+
+const Banner = styled.div`
+  background-image: url(${(props: BannerProps) => props.backgroundImageUrl || ''});
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:17]:
```suggestion
const Banner = styled.div<BannerProps>`
  background-image: url(${({ backgroundImageUrl }) => backgroundImageUrl || ''});
  background-position: center;
```

We can take advantage of [styled components custom props](https://styled-components.com/docs/api#using-custom-props)

`background-position: center;` keeps the banner vertically centred inside the div

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+
+const Logo = styled.img`
+  border-radius: 100%;
+  display: block;
+  margin-top: 20px;
+  width: 100px;
+`
+
+const ClubName = styled(PageTitle)`
+  margin-top: 30px;
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:62]: some clubs have horrendously long names, we should check what happens to them

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+const MemberCount = styled.div`
+  font-style: italic;
+  font-weight: ${fontWeight.bold};
+`
+
+const Categories = styled.div`
+  display: flex;
+`
+
+const Category = styled.p`
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:90]: multiline text inside a category is probably not desired, should add something to prevent it

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+  box-shadow: 0 5px 10px rgba(154, 160, 185, 0.05), 0 15px 100px rgba(166, 173, 201, 0.2);
+  display: flex;
+  margin-top: 10px;
+
+  @media ${device.mobileL} {
+    font-size: 10px;
+  }
+`
+
+const EventCardImg = styled.img`
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:170]: the image is getting stretched

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+            <EventCardImg src={e.backgroundImageURL} alt="" />
+            <EventCardDetails>
+              <EventCardTitle>{e.name}</EventCardTitle>
+              <EventCardDate>
+                {device.mobileS ? (
+                  <p>
+                    {e.startTime.format('dddd')}, {e.startTime.format('LL')}
+                    <br />
+                    {e.startTime.format('LT')} - {e.endTime.format('LT')} ET
+                  </p>
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:209]: somehow my laptop is getting the mobile case here...

is it possible to have a single format `e.startTime.format('dddd, LL')`, or is Moment not smart enough for that?

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
-      <p>{name}</p>
-      <p>{description}</p>
+      <EventsTitle>Upcoming Events</EventsTitle>
+      {upcomingEvents.map((e) => {
+        return <p>h</p>
+      })}
+      <PastEventsTitle>Past Events</PastEventsTitle>
+      {pastEvents.map((e) => {
+        return (
+          <EventCard>
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:199]: could factor this map function out into its own functional component

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
 
   return (
     <div>
-      <h1>Club Details</h1>
-      <p>{id}</p>
-      <p>{name}</p>
-      <p>{description}</p>
+      <EventsTitle>Upcoming Events</EventsTitle>
+      {upcomingEvents.map((e) => {
+        return <p>h</p>
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:194]: ℎ

xujustinj (changes requested):
overall structures lgtm, just a bunch of nits

client/pages/clubs/[id].tsx
Comment (ShreyasPrasad) [reply client/pages/clubs/[id].tsx:209]: According to https://momentjscom.readthedocs.io/en/latest/moment/04-displaying/01-format/, probably not.

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+        <SocialLink>techplusuw.ca</SocialLink>
+        <SocialLink style={{ marginLeft: '10px' }}>@techplusuw</SocialLink>
+      </Links>
+
+      <ClubMetaData>
+        <MemberCount>20-30 members</MemberCount>
+
+        {/* TODO: color code the tags */}
+        <Categories>
+          {tags.map((t) => (
```
Comment (ShreyasPrasad) [inline client/pages/clubs/[id].tsx:132]: would be nice to pull this into a small function for readability.

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+          ))}
+        </Categories>
+      </ClubMetaData>
+
+      <ClubDescription>{description}</ClubDescription>
+    </div>
+  )
+}
+
+const EventsTitle = styled.h2``
```
Comment (ShreyasPrasad) [inline client/pages/clubs/[id].tsx:143]: do we want to create empty styled components? super nitty, but I kinda see the value of both approaches (readability vs bloating codebase).

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+  margin: 0 20px 20px;
+  width: 100%;
+
+  @media ${device.mobileS} {
+    margin: 0 10px 10px;
+  }
+`
+
+const EventCardTitle = styled.p`
+  font-weight: 700;
```
Comment (ShreyasPrasad) [inline client/pages/clubs/[id].tsx:155]: is this a font-weight we want to standardize?

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+                  </p>
+                ) : (
+                  <p>
+                    {e.startTime.format('dddd')}, {e.startTime.format('LL')} |{' '}
+                    {e.startTime.format('LT')} - {e.endTime.format('LT')} ET
+                  </p>
+                )}
+              </EventCardDate>
+              <div style={{ display: 'flex', justifyContent: 'flex-end' }}>
+                {e.tags.map((t) => {
```
Comment (ShreyasPrasad) [inline client/pages/clubs/[id].tsx:218]: Again, for consistency it would be nice to render sub-components from another function cause we do that in other spots and it helps maintainability.

ShreyasPrasad (changes requested):
looks noice

client/pages/clubs/[id].tsx
Comment (xujustinj) [reply client/pages/clubs/[id].tsx:143]: I strongly prefer empty named styles. Adding customizations in the future is much easier if the style already exists

Comment (ShreyasPrasad) [reply client/pages/clubs/[id].tsx:143]: Yeh that makes sense.

client/context/Base.tsx
```diff
@@ -9,7 +9,7 @@ interface AppData {
   events: Map<Id, Event>
 }
 
-export type Club = {
+export interface Club {
```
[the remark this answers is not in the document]
jeffzh4ng: I personally am in the camp of using `type` as an alias (`type X = Y`) and using `interface` when defining a type. It's moreso choosing between the two depending on the semantic use case. Don't feel too strongly about it though, open to changing it back. Thoughts?

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+const Category = styled.p`
+  border: 1px solid black;
+  border-radius: 15px;
+  font-size: 12px;
+  margin-left: 5px;
+
+  padding: 5px 10px;
+  @media ${device.mobileS} {
+    padding: 2px 5px;
+  }
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:99]: Stumbled into this [thing](https://www.w3schools.com/css/css3_mediaqueries.asp) about responsive web design which says the default should be the mobile style

```suggestion
  padding: 2px 5px;
  @media not all and ${device.mobileS} {
    padding: 5px 10px;
  }
```

client/pages/clubs/[id].tsx
Comment (xujustinj) [reply client/pages/clubs/[id].tsx:97]: In that picture there happens to be no multiline text inside `Category` tags, I mean specifically preventing multiline `Category` tags.

All that said, we'll probably end up creating a shared `Tag` component and can defer until then.

P.S. the trick to avoiding multiline is a combination of
```css
  overflow: hidden;
  white-space: nowrap;
```
and setting a fixed `height` for the element

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,251 @@
...
+      <ClubName> {name}</ClubName>
+
+      <Links>
+        {linksThatExist.map((l) => (
+          <SocialLink>{l}</SocialLink>
+        ))}
+      </Links>
+
+      <ClubMetaData>
+        <MemberCount>20-30 members</MemberCount>
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:136]: uh we forgot to add this to our FE/BE Club schemas 😅 

cc @KTong821

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,251 @@
...
+    ))}
+  </CategoriesWrapper>
+)
+
+const EventsTitle = styled.h2`
+  margin-top: 32px;
+`
+
+const EventCardDetails = styled.div`
+  margin: 0 20px 20px;
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:159]: personally I prefer explicit margin-left/right/top/bottom instead of the shorthand, because I still haven't memorized the order

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,251 @@
...
+
+const EventCard = ({ event }: { event: BasicEvent }) => (
+  <EventCardWrapper>
+    <EventCardImg src={event.backgroundImageURL} alt="" />
+    <EventCardDetails>
+      <EventCardTitle>{event.name}</EventCardTitle>
+      <EventCardDate>
+        {device.mobileS ? (
+          <p>
+            {event.startTime.format('dddd')}, {event.startTime.format('LL')}
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:221]: `{event.startTime.format('dddd')}, {event.startTime.format('LL')}` this looks to be a constant that's used in both the mobile and large-screen case. Maybe we can factor this out into a string constant outside the JSX?

Likewise with `{event.startTime.format('LT')} - {event.endTime.format('LT')}`

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,251 @@
...
+            <br />
+            {event.startTime.format('LT')} - {event.endTime.format('LT')} ET
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:223]: hardcoded eastern time?

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,251 @@
...
+  margin: 0 20px 20px;
+  width: 100%;
+
+  @media ${device.mobileS} {
+    margin: 0 10px 10px;
+  }
+`
+
+const EventCardTitle = styled.p`
+  font-weight: ${weight700};
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:168]:
```suggestion
  font-weight: ${fontWeight.bold};
```

client/styles/typography.ts
```diff
@@ -16,10 +16,10 @@ const text12 = '12px'
...
 
 // font weight
-const weight900 = '900'
-const weight700 = '700'
-const weight600 = '600'
-const weight400 = '400'
+export const weight900 = '900'
+export const weight700 = '700'
+export const weight600 = '600'
+export const weight400 = '400'
```
xujustinj, line 22: I don't think this change is necessary, if we want to use `weight900` then we should add `superBold: weight900` to the `fontWeight` object already exported from here.

xujustinj (changes requested):

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx]: I'm against this approach, I think media queries applied to mobile make sense. Also w3schools shouldn't be the place where we go to for our source of truth.

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx:187]: fixed

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx:167]: Changed

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx]: Done

client/styles/typography.ts
jeffzh4ng: woops, you're right. reverted

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,251 @@
...
+const Container = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  margin: 0 auto 100px;
+  width: 65%;
+
+  @media ${device.mobileL} {
+    width: 85%;
+  }
```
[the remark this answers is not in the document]
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx:33]: I don't think mobile-first media queries read naturally. Why does percents only make sense on large screens? Mobile looks nice still as well.

image

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,251 @@
...
+  margin-bottom: 0;
+  margin-top: 32px;
+`
+
+const Links = styled.div`
+  display: flex;
+  margin-top: 16px;
+`
+
+const SocialLink = styled.div`
```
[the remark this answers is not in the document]
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx]: changed

client/pages/clubs/[id].tsx
```diff
@@ -210,35 +228,39 @@ const EventsHostedByClub = ({ events }: { events: Array<BasicEvent> }) => {
...
+  const date = `${event.startTime.format('dddd')}, ${event.startTime.format('LL')}`
+  const time = `${event.startTime.format('LT')} - ${event.endTime.format('LT')}`
+
+  const dateTimeString = (
+    <p>
+      {date}
+      {device.mobileS ? <br /> : ' '}
+      {time}
+    </p>
+  )
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:241]: beautiful

client/pages/clubs/[id].tsx
```diff
@@ -54,6 +55,7 @@ export default function ClubDetail() {
 
 const BackArrow = styled.img`
   margin-top: 64px;
+  cursor: pointer;
```
Comment (xujustinj) [inline client/pages/clubs/[id].tsx:58]: I'm quite unfamiliar with HTML/CSS but this kind of feels like a hack to make the image act like a link. Is it possible to do something like `<a><BackArrow ... /></a>` to make it automatically interpret the picture as a link?

Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx:58]: I believe there's a distinction being made in the webdev community between external links and in-app navigational components.

External links are actually links that link to websites outside of the current app. We should use `<a>` tags to give them the correct semantic meaning.

Internal navigational components like buttons to move forwards and backwards throughout the SPA arguably don't need `<a>` wrapped around itself. Thoughts?

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx:143]: @KTong821 we'll need to add member count to the backend APIs

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx]: @xujustinj thoughts?

client/pages/clubs/[id].tsx
Comment (xujustinj) [reply client/pages/clubs/[id].tsx:58]: Something like this? https://stackoverflow.com/questions/21715162/when-should-i-use-a-button-button-or-a-link-a-in-html

Makes sense, let's keep it the way it is then

client/pages/clubs/[id].tsx
Comment (xujustinj) [reply client/pages/clubs/[id].tsx]: Still a blog, but here's a more in-depth argument https://www.mightyminnow.com/2013/11/what-is-mobile-first-css-and-why-does-it-rock/

So the reason we would put the mobile style first is because that way it renders before the query's modifications. Especially on slower phones, this means there's no [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) that temporarily ruins the experience

client/pages/clubs/[id].tsx
```diff
@@ -18,6 +15,10 @@ const Banner = styled.div<BannerProps>`
   background-image: url(${({ backgroundImageUrl }) => backgroundImageUrl || ''});
   background-position: center;
   height: 30vh;
+
+  @media ${device.mobileS} {
+    height: 20vh;
+  }
```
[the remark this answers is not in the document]
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx]: Fixed

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx]: Changed CSS to be mobile first.

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx]: Added `moment-timezone`, all times are now adjust to the browser's local timezone, as well as the timezone stamp (EST/PST/etc)

image

xujustinj (approved):

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,228 @@
...
+
+  const club = clubs.get(parseInt(id))
+
+  return (
+    <div>
+      <Banner backgroundImageUrl={club.backgroundImageURL} />
+      <Container>
+        <ClubInfo club={club} />
+        <EventsHostedByClub events={club.events} />
+      </Container>
```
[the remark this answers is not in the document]
Comment (ShreyasPrasad) [reply client/pages/clubs/[id].tsx:57]: ^^^ :)

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,275 @@
...
+
+  return (
+    <div>
+      <BackArrow src="/back-arrow.svg" width="28px" onClick={() => router.back()} />
+
+      <Logo src={iconURL} alt="" />
+      <ClubName> {name}</ClubName>
+
+      <Links>
+        {linksThatExist.map((l) => (
```
Comment (ShreyasPrasad) [inline client/pages/clubs/[id].tsx:144]: might get JS warnings about rendered elements not containing keys here. Might be worthwhile to suppress or our console will get really red. I followed the convention of including dashed component name as prefix to any unique identifier: 
image

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,275 @@
...
+      </ClubMetaData>
+
+      <ClubDescription>{description}</ClubDescription>
+    </div>
+  )
+}
+
+const Categories = ({ tags }: { tags: Array<string> }) => (
+  <CategoriesWrapper>
+    {tags.map((t) => (
```
Comment (ShreyasPrasad) [inline client/pages/clubs/[id].tsx:162]: same comment as before with keys.

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,275 @@
...
+
+const EventsHostedByClub = ({ events }: { events: Array<BasicEvent> }) => {
+  const upcomingEvents = events.filter((e) => e.startTime.isAfter(moment()))
+
+  const pastEvents = events.filter((e) => e.startTime.isBefore(moment()))
+
+  return (
+    <div>
+      <EventsTitle>Upcoming Events</EventsTitle>
+      {upcomingEvents.map((e) => {
```
Comment (ShreyasPrasad) [inline client/pages/clubs/[id].tsx:220]: same comment about keys with these 2 maps.

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,275 @@
...
+  @media ${largerThan(mobile)} {
+    display: flex;
+    justify-content: flex-end;
+    margin-right: -4px;
+  }
+`
+
+const EventCategories = ({ tags }: { tags: Array<string> }) => (
+  <EventCategoriesWrapper>
+    {tags.map((t) => {
```
Comment (ShreyasPrasad) [inline client/pages/clubs/[id].tsx:271]: same comment about keys here too.

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,275 @@
...
+
+const Logo = styled.img`
+  border-radius: 100%;
+  display: block;
+  margin-top: 40px;
+  height: 100px;
+  width: 100px;
+`
+
+const ClubName = styled(PageTitle)`
```
Comment (ShreyasPrasad) [inline client/pages/clubs/[id].tsx:75]: TIL you can do this LOLL

client/pages/clubs/[id].tsx
```diff
@@ -1,22 +1,275 @@
...
+  margin-right: -4px;
+`
+
+const Category = styled.p`
+  border: 1px solid black;
+  border-radius: 15px;
+  margin: 0 4px;
+
+  padding: 4px 10px;
+  font-size: 14px;
```
Comment (ShreyasPrasad) [inline client/pages/clubs/[id].tsx:112]: really nit, but we should probably use the font size constants from` typography.ts` so if we change font sizes in the future it'll be dynamically changed.

ShreyasPrasad (changes requested):
lgtm, just a few small fixes

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx:144]: great catch

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx:162]: added keys

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx:220]: added keys

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx:271]: added keys

client/pages/clubs/[id].tsx
Comment (jeffzh4ng) [reply client/pages/clubs/[id].tsx]: great catch! i also changed another occurrence of hard-coded font size

ShreyasPrasad (approved):

xujustinj (approved):

client/pages/clubs/[id].tsx
Comment (xujustinj) [reply client/pages/clubs/[id].tsx:157]: created new issue #87 for this

PR #83, opened 2021-04-10
🗄 DB Population Script

KTong821:
## Purpose
V0 doesn't support creating new clubs/events yet, so this script does this by inserting those listed in a YAML file. 

## Testing
Insertions visible in DB; GET endpoints functioning.

## Method
Utilized Python library psycopg2 to connect/insert to Postgres server. YAML chosen for its easy-to-write syntax over JSON and supports nesting (unlike .csv).

## Notes
Requires db.ini and running

`pip install -r requirements.txt`

Add/remove clubs by changing `clubs.yml` file. Only 1-to-many relationship for clubs:events supported right now.

### DIFF
--- a/server/scripts/clubs.yml
+++ b/server/scripts/clubs.yml
@@ -1,0 +1,37 @@
+- name: Loolabs # All possible club fields listed
+  description: Connecting Waterloo students with projects
+  size: 10
+  banner_image: banner
+  icon_image: icon
+  facebook: facebook
+  twitter: twitter
+  instagram: instagram
+  website: website
+  events:
+    - name: Wednesday Work Session # All possible event fields listed
+      description: Building some stuff
+      url: github.com/loolabs/waterpark
+      banner_image: banner 1
+      start_time: 2021-01-01 00:00:00+00
+      end_time: 2021-01-01 00:00:00+00
+      facebook: facebook
+      twitter: twitter
+      instagram: instagram
+      tags:
+        - Engineering
+    - name: Saturday Work Session # All non-nullable event fields listed
+      description: Building more stuff
+      banner_image: banner 2
+      start_time: 2021-02-02 00:00:00+00
+      end_time: 2021-02-02 00:00:00+00
+  tags:
+    - Tech
+    - Community
+- name: Tech+ # All non-nullable club fields listed
+  description: >-
+    Tech+ Mentorship Program connects 1st- and 2nd-year UWaterloo students with experienced
+    and passionate upper-years from different domains in tech, and fosters this community
+    through events that bring everyone together.
+  size: 30
+  banner_image: banner
+  icon_image: icon
--- a/server/scripts/populate.py
+++ b/server/scripts/populate.py
@@ -6,14 +6,13 @@ from configparser import ConfigParser
 from datetime import datetime
 from os import path
 from enum import Enum
-from marshmallow import Schema, fields, validate
-import pprint
+
 import uuid
 import psycopg2
 import yaml
+from marshmallow import Schema, fields, validate
 
 filepath = "clubs.yml"
-pp = pprint.PrettyPrinter(indent=2)
 tag_entities = {}
 
 
--- a/server/scripts/populate.py
+++ b/server/scripts/populate.py
@@ -1,10 +1,12 @@
 import psycopg2
-import yaml  # noqa
+import yaml
 import uuid
 import pprint
 from datetime import datetime
 from os import path
 from configparser import ConfigParser
+from enum import Enum
+from marshmallow import Schema, fields, validate
 
 filepath = "clubs.yml"
 pp = pprint.PrettyPrinter(indent=2)
@@ -12,23 +14,54 @@ club_non_nullable = ["name", "description",
                      "size", "banner_image", "icon_image"]
 event_non_nullable = ["name", "description",
                       "banner_image", "start_time", "end_time"]
-tags_enum = ["Community", "Tech", "Creative", "Active", "Volunteering", "Gaming",
-             "Career", "Engineering", "Science", "Environment", "Arts", "Math", "Health"]
 tag_entities = {}
+
+
+class ClubSchema(Schema):
+    name = fields.Str(required=True, validate=validate.Length(min=1))
+    description = fields.Str(required=True, validate=validate.Length(min=1))
+    size = fields.Int(required=True, validate=validate.Range(min=1))
+    banner_image = fields.url(required=True, validate=validate.Length(min=1))
+    icon_image = fields.url(required=True, validate=validate.Length(min=1))
+
+
+class EventSchema(Schema):
+    name = fields.Str(required=True, validate=validate.Length(min=1))
+    description = fields.Str(required=True, validate=validate.Length(min=1))
+    banner_image = fields.url(required=True, validate=validate.Length(min=1))
+    start_time = fields.DateTime(required=True)
+    end_time = fields.DateTime(required=True)
+
+
+class Tags(str, Enum):
+    community = "Community"
+    tech = "Tech"
+    creative = "Creative"
+    active = "Active"
+    volunteering = "Volunteering"
+    gaming = "Gaming"
+    career = "Career"
+    engineering = "Engineering"
+    science = "Science"
+    environment = "Environment"
+    arts = "Arts"
+    math = "Math"
+    health = "Health"
+
 
 # Assumptions: each club has its own events. Many-to-many not supported
 # yet. See example clubs.yml file for example input. db.ini file must be
 # in same directory
 
-def insert_tags(): 
+def insert_tags():
     conn = connect()
     try:
         with conn.cursor() as cursor:
-            for tag in tags_enum:
+            for tag in Tags:
                 _id = str(uuid.uuid4())
-                tag_entities[tag] = _id
+                tag_entities[tag.value] = _id
                 cursor.execute("INSERT INTO tag VALUES (%s,%s,%s,%s);",
-                               (_id, datetime.now(), datetime.now(), tag))
+                               (_id, datetime.now(), datetime.now(), tag.value))
         conn.commit()
         print("Tags Inserted")
     finally:
@@ -52,7 +85,8 @@ def read_clubs(filepath):
 def verify(entity, fields, typeof):
     for field in fields:
         if entity.get(field) is None:
-            name = "Untitled" if entity.get("name") is None else entity.get("name")
+            name = "Untitled" if entity.get(
+                "name") is None else entity.get("name")
             print("Error:", typeof, name,
                   "is missing mandatory field", field)
             return False
@@ -76,10 +110,7 @@ def insert_clubs():
     if club_list is None:
         print("Something failed; aborting...")
         return
-    conn = connect()
-    if conn is None:
-        return
-    insert_tags()
+    
     try:
         while True:
             club = next(club_list)
@@ -129,7 +160,7 @@ def connect():
         return None
     parser = ConfigParser()
     parser.read("db.ini")
-    try:        
+    try:
         conn = psycopg2.connect(
             host=parser["postgresql"].get("host"),
             database=parser["postgresql"].get("database"),
@@ -147,4 +178,7 @@ def connect():
             print("Connection Closed")
 
 
+conn = connect()
+
+insert_tags()
 insert_clubs()
--- a/server/scripts/populate.py
+++ b/server/scripts/populate.py
@@ -1,0 +1,150 @@
+import psycopg2
+import yaml  # noqa
+import uuid
+import pprint
+from datetime import datetime
+from os import path
+from configparser import ConfigParser
+
+filepath = "clubs.yml"
+pp = pprint.PrettyPrinter(indent=2)
+club_non_nullable = ["name", "description",
+                     "size", "banner_image", "icon_image"]
+event_non_nullable = ["name", "description",
+                      "banner_image", "start_time", "end_time"]
+tags_enum = ["Community", "Tech", "Creative", "Active", "Volunteering", "Gaming",
+             "Career", "Engineering", "Science", "Environment", "Arts", "Math", "Health"]
+tag_entities = {}
+
+# Assumptions: each club has its own events. Many-to-many not supported
+# yet. See example clubs.yml file for example input. db.ini file must be
+# in same directory
+
+def insert_tags(): 
+    conn = connect()
+    try:
+        with conn.cursor() as cursor:
+            for tag in tags_enum:
+                _id = str(uuid.uuid4())
+                tag_entities[tag] = _id
+                cursor.execute("INSERT INTO tag VALUES (%s,%s,%s,%s);",
+                               (_id, datetime.now(), datetime.now(), tag))
+        conn.commit()
+        print("Tags Inserted")
+    finally:
+        conn.close()
+
+
+def club_generator(clubs):
+    for club in clubs:
+        yield club
+
+
+def read_clubs(filepath):
+    with open(filepath) as file:
+        clubList = yaml.load(file, Loader=yaml.FullLoader)
+        for club in clubList:
+            if not verify(club, club_non_nullable, "club"):
+                return None
+        return club_generator(clubList)
+
+
+def verify(entity, fields, typeof):
+    for field in fields:
+        if entity.get(field) is None:
+            name = "Untitled" if entity.get("name") is None else entity.get("name")
+            print("Error:", typeof, name,
+                  "is missing mandatory field", field)
+            return False
+        events = entity.get("events")
+        if events is not None:
+            for event in events:
+                if not verify(event, event_non_nullable, "event"):
+                    return False
+        tags = entity.get("tags")
+        if tags is not None:
+            for tag in tags:
+                if tag not in tags_enum:
+                    print("Error:", typeof, entity.get(
+                        "name"), "has illegal tag", tag)
+                    return False
+    return True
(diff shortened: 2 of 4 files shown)


server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+                                cursor.execute(
+                                    "INSERT INTO tag_events VALUES (%s, %s);", (tag_entities.get(tag), event.get("id")))
+                tags = club.get("tags")
+                if tags is not None:
+                    for tag in tags:
+                        cursor.execute(
+                            "INSERT INTO tag_clubs VALUES (%s, %s);", (tag_entities.get(tag), club.get("id")))
+    except StopIteration:
+        conn.commit()
+        print("Clubs Inserted")
```
xujustinj, line 118: The usual `for` loop syntax is fine here

```suggestion
    for club in club_list:
        club["id"] = str(uuid.uuid4())
        # pp.pprint(club)
        with conn.cursor() as cursor:
            cursor.execute("INSERT INTO club VALUES (%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s);",
                           (club.get("id"), datetime.now(), datetime.now(), club.get("name"),
                            club.get("description"), club.get("size"), club.get(
                                "banner_image"), club.get("icon_image"),
                            club.get("facebook"), club.get("twitter"), club.get("instagram"), club.get("website")))
            events = club.get("events")
            if (events is not None):
                for event in events:
                    event["id"] = str(uuid.uuid4())
                    cursor.execute(
                        "INSERT INTO event VALUES (%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s);",
                        (event.get("id"), datetime.now(), datetime.now(), event.get("name"), event.get("description"),
                         event.get("url"), event.get("banner_image"),
                         event.get("start_time"), event.get("end_time"),
                         event.get("facebook"), event.get("twitter"), event.get("instagram")))
                    cursor.execute(
                        "INSERT INTO club_events VALUES (%s, %s);", (club.get("id"), event.get("id")))
                    tags = event.get("tags")
                    if tags is not None:
                        for tag in tags:
                            cursor.execute(
                                "INSERT INTO tag_events VALUES (%s, %s);", (tag_entities.get(tag), event.get("id")))
            tags = club.get("tags")
            if tags is not None:
                for tag in tags:
                    cursor.execute(
                        "INSERT INTO tag_clubs VALUES (%s, %s);", (tag_entities.get(tag), club.get("id")))
```

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+                    return False
+    return True
+
+
+def insert_clubs():
+    club_list = read_clubs(filepath)
+    if club_list is None:
+        print("Something failed; aborting...")
+        return
+    conn = connect()
```
xujustinj, line 79: This is fine in our limited use case, but in general we want to avoid calling potentially-expensive functions like `connect` as much as possible.

`connect` is called in both `insert_tags` and `insert_clubs` to get access to the same database both times. Is it possible for the two functions to share the same connection?

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+from configparser import ConfigParser
+
+filepath = "clubs.yml"
+pp = pprint.PrettyPrinter(indent=2)
+club_non_nullable = ["name", "description",
+                     "size", "banner_image", "icon_image"]
+event_non_nullable = ["name", "description",
+                      "banner_image", "start_time", "end_time"]
+tags_enum = ["Community", "Tech", "Creative", "Active", "Volunteering", "Gaming",
+             "Career", "Engineering", "Science", "Environment", "Arts", "Math", "Health"]
```
xujustinj, line 16: Python has actual `Enum`s that are much better suited for this.

```py
class Tags(str, Enum):
    community = "Community"
    # and so on
```

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+        return
+    conn = connect()
+    if conn is None:
+        return
+    insert_tags()
+    try:
+        while True:
+            club = next(club_list)
+            club["id"] = str(uuid.uuid4())
+            # pp.pprint(club)
```
xujustinj, line 87: check out [`logging.debug`](https://docs.python.org/3/library/logging.html#logging.Logger.debug) instead

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+            with conn.cursor() as cursor:
+                cursor.execute("INSERT INTO club VALUES (%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s);",
+                               (club.get("id"), datetime.now(), datetime.now(), club.get("name"),
+                                club.get("description"), club.get("size"), club.get(
+                                    "banner_image"), club.get("icon_image"),
+                                club.get("facebook"), club.get("twitter"), club.get("instagram"), club.get("website")))
+                events = club.get("events")
+                if (events is not None):
+                    for event in events:
```
xujustinj, line 96: Python lets you call `.get` with a default value to return if the key does not exist. This gives us the one-liner

```py
for event in club.get("events", []):
```

If you end up turning clubs into Python objects, then this probably won't be necessary (you can default the field to the empty list, so that it is never `None`), and then `for event in club.events` is enough

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+                             event.get("url"), event.get("banner_image"),
+                             event.get("start_time"), event.get("end_time"),
+                             event.get("facebook"), event.get("twitter"), event.get("instagram")))
+                        cursor.execute(
+                            "INSERT INTO club_events VALUES (%s, %s);", (club.get("id"), event.get("id")))
+                        tags = event.get("tags")
+                        if tags is not None:
+                            for tag in tags:
+                                cursor.execute(
+                                    "INSERT INTO tag_events VALUES (%s, %s);", (tag_entities.get(tag), event.get("id")))
```
xujustinj, line 110: This is a "request in a loop", which could cause some serious performance (and thus finance) issues. The resolution is to do as much as possible in a single `execute` statement. There seems to be a variety of ways to do this, according to [this SO thread](https://stackoverflow.com/questions/8134602/psycopg2-insert-multiple-rows-with-one-query)

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+
+def insert_clubs():
+    club_list = read_clubs(filepath)
+    if club_list is None:
+        print("Something failed; aborting...")
+        return
+    conn = connect()
+    if conn is None:
+        return
+    insert_tags()
```
xujustinj, line 82: it probably makes sense to call this separately from `insert_clubs` by adding another statement at the end of the file

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+        print("Missing configuration file.")
+        return None
+    parser = ConfigParser()
+    parser.read("db.ini")
+    try:        
+        conn = psycopg2.connect(
+            host=parser["postgresql"].get("host"),
+            database=parser["postgresql"].get("database"),
+            user=parser["postgresql"].get("user"),
+            password=parser["postgresql"].get("password"))
```
xujustinj, line 137: Accessing a dictionary by string key is actually quite slow. `parser["postgresql"]` should be stored in an intermediate variable so that it's only called once

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+        print("Connection Successful")
+        cursor = conn.cursor()
+        cursor.execute("SELECT version()")
+        print(cursor.fetchone())
+        return conn
+    except (Exception, psycopg2.DatabaseError) as error:
+        print(error)
+        if conn is not None:
```
xujustinj, line 145:
```suggestion
    finally:
        if conn is not None:
```

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def connect():
+    conn = None
+    print("Connecting to PostgreSQL server...")
+    if not path.exists("db.ini"):
+        print("Missing configuration file.")
+        return None
+    parser = ConfigParser()
+    parser.read("db.ini")
```
xujustinj, line 131: Since it's developers rather than normal people who will use this script, it's probably more useful to raise an exception and fail in a noisy way (which comes with a useful stack trace and all).

The best way to raise an exception is to do something that will fail if your condition is not met. This is part of the [EAFP](https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/) idiom in Python

```suggestion
    with open("db.ini") as f:
        parser.read_file(f)
```

xujustinj (changes requested):

server/scripts/clubs.yml
```diff
@@ -0,0 +1,37 @@
+- name: Loolabs # All possible club fields listed
```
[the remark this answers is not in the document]
KTong821: I mean, most info should be public anyways so I don't see why not.

server/scripts/requirements.txt
```diff
@@ -0,0 +1,11 @@
+Package     Version
```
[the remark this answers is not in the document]
KTong821: I think I did `pip list > requirements.txt`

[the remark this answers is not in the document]
KTong821: I just realized it should've been `pip freeze` instead.

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
+import psycopg2
+import yaml  # noqa
```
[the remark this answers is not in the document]
KTong821: I use Anaconda, and it was complaining about not being able to find yaml. It ran perfectly well, so I silenced it.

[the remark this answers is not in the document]
xujustinj: According to conda issue 5142 (which I will not link here to avoid showing up in their issue), you might have to `conda install` instead of `pip install`

server/scripts/clubs.yml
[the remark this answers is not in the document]
jeffzh4ng: I don't think there's any harm in pushing club data but we probably make a team-wide decision for all future population scripts. In the future, there might be data that must be private and hence can't be pushed to version control. It would then be confusing that some data is on github and some not.

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+import uuid
+import pprint
+from datetime import datetime
+from os import path
+from configparser import ConfigParser
+
+filepath = "clubs.yml"
```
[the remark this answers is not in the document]
jeffzh4ng: @xujustinj like a cli?

jeffzh4ng (dismissed):
Looks good to me, don't have too much changes to suggest as I don't have much Python experience. When Justin gives the approval then lgtm.

server/scripts/populate.py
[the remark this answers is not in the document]
xujustinj: you can use `argparse` to specify a filepath argument

server/scripts/populate.py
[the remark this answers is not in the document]
KTong821: Fixed; vscode issue.

server/scripts/populate.py
KTong821: I think I used pretty print for the sake of readability. I can get rid of the comment though.

server/scripts/populate.py
```diff
@@ -0,0 +1,150 @@
...
+                if not verify(event, event_non_nullable, "event"):
+                    return False
+        tags = entity.get("tags")
+        if tags is not None:
+            for tag in tags:
+                if tag not in tags_enum:
+                    print("Error:", typeof, entity.get(
+                        "name"), "has illegal tag", tag)
+                    return False
+    return True
```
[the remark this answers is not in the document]
KTong821: Implemented in latest commit. One issue though: marshmallow doesn't tell you which object specifically the error comes from, it just says that xyz field is missing. I could add a simple print statement, but that could clutter output.

server/scripts/populate.py
```diff
@@ -1,138 +1,181 @@
+# Assumptions: each club has its own events. Many-to-many not supported
+# yet. See example clubs.yml file for example input. db.ini file must be
+# in same directory
+
 from configparser import ConfigParser
 from datetime import datetime
 from os import path
+from enum import Enum
+from marshmallow import Schema, fields, validate
```
xujustinj, line 9: super nit: I think this import might be in the wrong place

server/scripts/populate.py
```diff
@@ -1,138 +1,181 @@
...
+    events = fields.List(fields.Nested(EventSchema))
+    tags = fields.List(fields.Str(
+        validate=validate.OneOf([tag.value for tag in Tags])))
+
+
+def insert_tags():
+    tag_args = []
+    for tag in Tags:
+        _id = str(uuid.uuid4())
+        tag_entities[tag.value] = _id
```
xujustinj, line 73: 🧙‍♀️
For even more cursed python magic, you can actually bake the ID right into the `Tags` enum, so that the `tag_entities` dictionary won't be necessary anymore.

(That's definitely optional though)

server/scripts/populate.py
```diff
@@ -1,138 +1,181 @@
...
+    career = "Career"
+    engineering = "Engineering"
+    science = "Science"
+    environment = "Environment"
+    arts = "Arts"
+    math = "Math"
+    health = "Health"
+
+
+class EventSchema(Schema):
```
xujustinj, line 40: so clean 😍

server/scripts/populate.py
```diff
@@ -148,4 +167,14 @@ def connect():
             print("Connection Closed")
 
 
+conn = connect()
+
+if conn is None:
+    exit()
+
+print(Tags.active.value)
```
xujustinj, line 175:
```suggestion
```

server/scripts/populate.py
```diff
@@ -0,0 +1,180 @@
...
+
+    club_values = []
+    event_values = []
+    club_events_values = []
+    tag_events_values = []
+    tag_clubs_values = []
+
+    for club in club_list:
+        club["id"] = str(uuid.uuid4())
+        club_values.append((club.get("id"), datetime.now(), datetime.now(), club.get("name"),
```
xujustinj, line 106: Since we're now loading clubs using the schema, I'm pretty sure `club` is not a dictionary here, so it doesn't make sense to add new properties to it. Correct me if I'm wrong.

We should probably do this instead:
```suggestion
        clubId = str(uuid.uuid4())
        club_values.append((clubId, datetime.now(), datetime.now(), club.get("name"),
```

server/scripts/populate.py
```diff
@@ -0,0 +1,180 @@
...
+                            club.get("description"), club.get("size"), club.get(
+                                "banner_image"), club.get("icon_image"),
+                            club.get("facebook"), club.get("twitter"), club.get("instagram"), club.get("website")))
```
xujustinj, line 109: likewise, lines like this can now be
```suggestion
                            club.facebook, club.twitter, club.instagram, club.website))
```

xujustinj (changes requested):
I'm not sure the current `insert_clubs` actually works correctly, is the latest code tested?

server/scripts/populate.py
KTong821: I think .load() returns a dictionary by default. If we want an object, then there's an extra method that needs to be declared in a new Club class. See [here](https://marshmallow.readthedocs.io/en/stable/quickstart.html#deserializing-objects-loading). I'm not sure if Python's built-in dict->object is any more efficient than what I have here. Would like to hear what you think.

server/scripts/populate.py
KTong821: Not sure what you mean...things in the wrong order?

xujustinj: yeah, I think it should be after pprint/uuid

server/scripts/populate.py
```diff
@@ -0,0 +1,179 @@
...
+from os import path
+from enum import Enum
+from marshmallow import Schema, fields, validate
+import pprint
+import uuid
+import psycopg2
+import yaml
+
+filepath = "clubs.yml"
+pp = pprint.PrettyPrinter(indent=2)
```
xujustinj, line 16: doesn't look like you're using this anymore, you can remove

```suggestion
```

also can remove the pprint import

server/scripts/populate.py
xujustinj: There are tools like [`dataclasses-json`](https://pypi.org/project/dataclasses-json/) to cover the last step of converting to objects, but that would require redoing the schema and everything. What you have is fine for now

server/scripts/populate.py
```diff
@@ -0,0 +1,179 @@
...
+        print("Something failed; aborting...")
+        return
+
+    club_values = []
+    event_values = []
+    club_events_values = []
+    tag_events_values = []
+    tag_clubs_values = []
+
+    for club in club_list:
```
xujustinj, line 104: `read_clubs` can no longer return `None`

```suggestion
    club_values = []
    event_values = []
    club_events_values = []
    tag_events_values = []
    tag_clubs_values = []

    for club in read_clubs(filepath):
```

xujustinj (changes requested):
Great work! Just a few more things

server/scripts/populate.py
KTong821: Like so?

`from configparser import ConfigParser
from datetime import datetime
from os import path
from enum import Enum

import uuid
import psycopg2
import yaml

from marshmallow import Schema, fields, validate`

xujustinj (approved):
🚀

jeffzh4ng (approved):

PR #84, opened 2021-04-26
🗄 Fix production database

jeffzh4ng:
## Screenshots
image
Deploying our backend server on Heroku now works!

## Purpose
Fixes our backend crashing when deploying to Heroku.

## Approach
Fixes production database by turning SSL connections on.

## Testing
Manually verified.

## Learning
Heroku Postgres supports SSL connections by default.


xujustinj (approved):
we paired on this and tested it together

PR #104, opened 2021-05-14
🗄 Enable CORS

jeffzh4ng:
## Purpose
Enables CORS on our backend.

## Approach
There's two approaches to enabling cors with Express.

The first approach is use the package `cors`.

```
import express from 'express'
import cors from 'cors'

const app = express()
app.use(cors())

```

The second approach is to forgo any external dependency, and enable CORS manually.

```
import express from 'express'
import cors from 'cors'

const enableCorsMiddleware = (_req: Request, res: Response, next: NextFunction) => {
  res.header('Access-Control-Allow-Origin', '*')
  res.header(
    'Access-Control-Allow-Headers',
    'Origin,X-Requested-With,Content-Type,Accept,content-type,application/json'
  )
  next()
}

const app = express()
app.use(enableCorsMiddleware)

```

I tried the first approach because it was simpler albeit less efficient. For some weird reason node is having a hard time finding the `cors` module.

The PR opts for the second approach.

I'm too lazy to look into what's wrong with the npm and the `cors` package, and we need to enable cors asap (tmr is Spring BOT) so our frontend requests don't fail. If someone is completely against approach 2 I'm happy to look into fixing approach 1, but honestly I think we need to move fast and focus on product instead of engineering this term ;)


KTong821 (dismissed):
LGTM, but shouldn't this come after Justin's IoC PR?

xujustinj (approved):
LGTM, although option 1 is definitely more ideal

Some more detail on Node not being able to find `cors` would be nice. One way it might happen is if you `npm install cors` locally but forget to run `docker-compose up --build`

xujustinj:
> LGTM, but shouldn't this come after Justin's IoC PR?

Doesn't matter too much, since it's a small change. The compatibility changes for IOC would be a simple matter of copying this code into one of the setup functions.

server/package.json
```diff
@@ -20,7 +20,9 @@
     "@mikro-orm/postgresql": "^4.3.4",
     "@mikro-orm/reflection": "^4.3.4",
     "@mikro-orm/sql-highlighter": "^1.0.1",
+    "@types/cors": "^2.8.10",
```
xujustinj, line 23: I think this should go in `devDependencies`

server/src/index.ts
```diff
@@ -26,7 +27,9 @@ const initializeORM = async (DBObject: typeof DB) => {
 
 const bootstrap = async () => {
   await initializeORM(DB)
+
   app.use(express.json())
+  app.use(cors())
```
xujustinj, line 32: you might want to rebase, now that #70 is in

xujustinj (changes requested):
Someday, I gotta set up the npm workspaces properly, it's not sustainable to keep telling people to delete `server/package-lock.json` and re-run `npm install` from the root folder

That said, I think you need to delete `server/package-lock.json` and re-run `npm install` from the root folder

jeffzh4ng:
@xujustinj @KTong821 I'm a 🤡 , completely forgot I had to run `docker-compose up --build`

docker-compose.yml
```diff
@@ -12,8 +12,8 @@ services:
     ports:
       - ${SERVER_PORT}:${SERVER_PORT}
     # command: npm run test:unit
-    command: npm run test:integration
-    # command: npm run start:dev
+    # command: npm run test:integration
+    command: npm run start:dev
```
xujustinj, line 16: thanks for the catch!

xujustinj (approved):
we might be missing a change to the root `package-lock.json` here but it's not like npm respects that file anyway

xujustinj (approved):
🚀

jeffzh4ng:
@xujustinj Updated the root `package-lock.json`. Do we have notion docs for how npm workspaces work and the process in adding a package to Waterpark?

xujustinj:
> @xujustinj Updated the root `package-lock.json`. Do we have notion docs for how npm workspaces work and the process in adding a package to Waterpark?

That's in my todo list for today!

xujustinj (approved):
🤷

jeffzh4ng:
Updating the root `package-lock.json` causes a unit tests to fail on GitHub Actions CI. Omitting the update of `package-lock.json` for now.

```
npm ERR! enoent ENOENT: no such file or directory, chmod '/home/runner/work/waterpark/waterpark/node_modules/npm/node_modules/@npmcli/installed-package-contents/index.js'
npm ERR! enoent This is related to npm not being able to find a file.
```

KTong821 (approved):
LGTM, actually implemented the same thing at coop yesterday.

PR #105, opened 2021-05-14
💻  Fetch live clubs and events with React Query

jeffzh4ng:
## Screenshots
image

image

There are no images because our API doesn't have real images right [now](https://waterpark-staging.herokuapp.com/api/v1/clubs).

## Purpose
Fetches live clubs and events with React Query.

## Approach
The PR installs React Query as our frontend's http client. Check out the spike in Notion for more details on why React Query was chosen over SWR.

Creates a new folder until utils called `serde` (short for *ser*ialize and *de*seerialize). Another potential name for this folder can simply be `api`. This folder contains deserializers that convert DTO's to client-side types.


client/context/Base.tsx
```diff
@@ -13,12 +16,14 @@ export interface Club {
...
-  instagramLink?: string
-  websiteLink?: string
+  links: {
+    bannerImage: string
+    iconImage: string
+    facebook?: string
+    twitter?: string
+    instagram?: string
+    website?: string
+  }
```
jeffzh4ng, line 26: Changed frontend type to mirror backend type.

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.8"
```
xujustinj, line 1: I feel like this isn't the first time we've seen these quotation marks change. Do we have conflicting Prettier setups somewhere?

client/utils/serde/clubs-and-events.ts
```diff
@@ -0,0 +1,82 @@
...
+  eventId: number,
+  club: ClubDTO
+): Event => {
+  const localZone = Intl.DateTimeFormat().resolvedOptions().timeZone
+
+  return {
+    ...serializedEvent,
+    id: eventId,
+    startTime: moment.tz('2022-11-18 10:55', 'Asia/Tokyo').local().tz(localZone),
+    endTime: moment.tz('2022-11-18 10:55', 'Asia/Tokyo').local().tz(localZone),
```
xujustinj, line 76: Still hard-coding for now? I guess a future todo is to deserialize the actual date.

client/utils/serde/clubs-and-events.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  let eventId = 0
+  const deserializedEvents: Array<Event> = []
+
+  const deserializedClubs = serializedClubs.map((club, clubId) => {
+    return {
+      ...club,
+      id: clubId,
+      events: club.events.map((event) => {
+        const deserializedEvent = deserializeEvent(event, eventId++, club)
```
xujustinj, line 51: so `eventId` seems to always be equal to `deserializedEvents.length`, maybe we can just use that?
```suggestion
        const deserializedEvent = deserializeEvent(event, deserializedEvents.length, club)
```

xujustinj (changes requested):
Need to fix the `client/package-lock` again, sorry.

I'm not seeing any code that uses React Query to fetch from the backend, did I miss something?

client/components/club-details/ClubInfo.tsx
```diff
@@ -63,15 +63,15 @@ interface ClubInfoProps {
 export const ClubInfo = ({ club }: ClubInfoProps) => {
   const router = useRouter()
 
-  const { name, description, iconURL, tags } = club
-  const links = [club.facebookLink, club.twitterLink, club.instagramLink, club.websiteLink]
-  const linksThatExist = links.filter((l) => l !== undefined)
+  const { name, description, links, tags } = club
+  const potentialLinks = [links.facebook, links.twitter, links.instagram, links.website]
+  const linksThatExist = potentialLinks.filter((l) => l !== undefined)
```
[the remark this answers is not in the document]
jeffzh4ng: Changed.

docker-compose.yml
jeffzh4ng: Someone must be changing them to double, but we should be using single quotes.

client/utils/serde/clubs-and-events.ts
jeffzh4ng: you're a smart cs mans justin

client/utils/serde/clubs-and-events.ts
jeffzh4ng: woops, fixed. dates are working correctly :D

client/utils/serde/clubs-and-events.ts
```diff
@@ -0,0 +1,81 @@
...
+  eventId: number,
+  club: ClubDTO
+): Event => {
+  const localZone = Intl.DateTimeFormat().resolvedOptions().timeZone
+
+  return {
+    ...serializedEvent,
+    id: eventId,
+    startTime: moment.tz(serializedEvent.startTime, 'Asia/Tokyo').local().tz(localZone),
+    endTime: moment.tz(serializedEvent.endTime, 'Asia/Tokyo').local().tz(localZone),
```
xujustinj, line 75: I have no idea how moment timezones work (timezone documentation is unreadable to me), but I think we should either pass no timezone at all or GMT

xujustinj (approved):

client/utils/serde/clubs-and-events.ts
ShreyasPrasad: Yeh, I think that was solely for testing.

ShreyasPrasad (approved):

client/utils/serde/clubs-and-events.ts
jeffzh4ng: This timezone string (data) should actually come from the backend. I can change this client code to 'America/Toronto' on the client and we assume every time from the backend is EST.

Let's merge this in today before Spring BOT and then have a discussion later on.

Issue #124, opened 2021-05-22
Change pointer cursor when you're hovering over a link


evtyz:
On my device, when you hover over a card for clubs or events, there's no indication that the card is clickable and links to an additional page. (Exception: hovering over images on events.) We should probably change the cursor to become a pointer when the user hovers these cards, so it's obvious you can click them.

Feel free to dismiss this issue if it's intentional.

PR #126, opened 2021-05-22
💻 Change cursor to pointer on event/club cards

evtyz:
## Purpose
Resolves https://github.com/loolabs/waterpark/issues/124
In the UI cards for event list view and club view, the cursor was not set to "pointer" status even though the cards were clickable. 

## Approach
The components that had an `onClick` attribute had the `cursor: pointer;` CSS property added.

## Testing
Tested via visual inspection.


xujustinj (approved):
nice catch 🎉

wang-em (approved):
lgtm

PR #128, opened 2021-05-23
🌍 Remove package-lock.json

KTong821:
## Purpose
The `package-lock.json` file turns out to be an OS and npm version dependent file, which frequently causes headaches when merging. This PR deletes this file and adds it to the .gitignore.


xujustinj (approved):
If we ever change our minds about this, it's really easy to bring back the file

Issue #134, opened 2021-05-24
Rewrite calendar component in Calendar View


wang-em:
The calendar component should be rewritten to match the design in figma. Currently the scroll bar appears at weird places and resizing the calendar does not match the ListView title. 
We decided to keep the mobile view for now (3 days view), but further investigations are being made for the design

wang-em:
Instead of rewriting the entire component we propose the following changes:
1. Hide calendar view on mobile
2. Add ability to go to prev and next day on mobile
3. Add styling for event cards on desktop
4. Add ability to go to prev and next week on desktop
5. Fix margin at breakpoints to prevent weird positioning and scroll bar

PR #137, opened 2021-05-28
💻 Add header to events page

### LINKED ISSUE #121: Implement header for events page
Implement a similar header for the events page as the one that already exists for the clubs page.

Figma reference:
https://www.figma.com/file/d7RH1W7Hjj09dT5ZoENqPO/Waterpark---V0?node-id=597%3A2518

xujustinj:
## Screenshots

https://youtu.be/NKigUdlXkpg

## Purpose

Closes #121 

Also hook up event search to the events page. At the moment this performs filtering before everything else (including the creation of the map from days to events). This may or may not be ideal, pending future discussion.

## Approach

Reuses the same header components from the clubs page, but duplicates the code. If we were to properly turn the header into a common component, the refactor could break a bunch of changes in the pipeline. Some future refactor is needed.

## Testing

Visually verified in screencap

### DIFF
--- a/client/components/SearchInput.tsx
+++ b/client/components/SearchInput.tsx
@@ -1,5 +1,5 @@
 import styled from 'styled-components'
-import { colours, fontWeight, fontInter } from '../../styles'
+import { colours, fontWeight, fontInter } from '../styles'
 
 const mobile = `425px`
 const tablet = `768px`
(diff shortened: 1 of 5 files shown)


jeffzh4ng (approved):
For now duplicating the header is ok to get this bug fixed. A refactor would be nice though in the future.

wang-em (approved):

xujustinj:
Discussed with @wang-em, it makes sense to merge this now so that we have a header to use for reference when we do the calendar fixes for #134

PR #145, opened 2021-06-12
💻 Update frontend README with stuff from Notion

xujustinj:
As we decided in the last retro, things specific to Waterpark's implementation should be moved into Waterpark READMEs, rather than our Notion which is for organization-level stuff. This PR makes the change for FE.

Sidenote: I'm not sure if the stuff added to the FE README here should be straight-up deleted from Notion. For example, the tech stack stuff is still relevant, since the Waterpark stack is a subset of the Loo Labs stack.


evtyz (approved):
LGTM!

a22yuen (approved):
LGTM!!

PR #157, opened 2021-06-19
132/html lang en

### LINKED ISSUE #132: Add HTML lang attribute for accessibility
This helps with assistive reading (text-to-speech) if we mention that the language is English in the code. (something like `<html lang="en">` ?) Not really a priority but nice-to-have.

a22yuen:
## Screenshot
Before
image
After
image

## Purpose
Close #132 
Add en to lang attribute of html tag for assistive reading

## Approach
Used the next.config.js to change the tag. An alternative would be to change the _document.tsx, but that requires changing its template code. Doing it this way is more straightforward and isolates the change to the next.config file

## Testing
_How did you test your changes?_
Checked the page before and after adding changes to ensure the lang=en in html tag

#### Blog Posts
https://melvingeorge.me/blog/set-html-lang-attribute-in-nextjs


client/next.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  i18n: {
+    locales: ['en-US'],
+    defaultLocale: 'en-US',
+  },
+}
```
xujustinj, line 6:
```suggestion
}

```

xujustinj (approved):

wang-em (approved):

PR #159, opened 2021-06-19
🗄️ Add EventArgs types to AfterCreate param

### LINKED ISSUE #156: Look into Mikro's AfterCreate parameter types
See `TODO: fix any type` comments in `event.entity.ts`, `club.entity.ts`, and `user.entity.ts`. There should be a type for the `target` parameter, but we don't know what it should be... see the discussion in https://github.com/loolabs/auth/pull/2#discussion_r650405323 and https://github.com/loolabs/auth/pull/2#discussion_r650410515

If we cannot determine it, set the type to `unknown`.

KTong821:
## Purpose
Resolves #156. Type definitions can be found [here](https://mikro-orm.io/docs/lifecycle-hooks/#eventsubscriber)


xujustinj (approved):
Nice find! Let's make the same changes in the auth repo

jeffzh4ng (approved):
🙌

PR #173, opened 2021-07-12
💻 Make navbar responsive on mobile

evtyz:
### Before:
image
#### Tablet:
image
#### Phone: (This is animated but the screenshot is still)
image

## Purpose
The navbar was not responsive on devices with smaller screen width. This closes #172.

## Approach
Media queries to reduce margin and font size were used for tablets.
I created another div for the mobile dropdown and some CSS transitions for phones.

## Testing
Via visual inspection.

PR #174, opened 2021-07-12
💻 Refactor media breakpoints

### LINKED ISSUE #171: Clean up media queries in the front-end
Some components are using breakpoints defined in `styles/device.ts`, some are doing their own thing. Find the right people, and come up with a solution to consolidate it all.

evtyz:
## Purpose
Closes #171.

## Approach
Previously, many files had their own const strings for `mobile`, `tablet`, and `laptop`. They also had their own `smallerThan()` and `largerThan()` functions.

Also, the main file that was supposed to regulate breakpoints, `styles/device.ts`, was not organized well. It exported a `device` object that was supposed to give templates for different breakpoints, but did not cover all possible use cases and was extremely inconsistent. E.g. `laptop` would be `max-width` and `desktop` would be `min-width` for whatever reason. Apparently people got sick of this and just started using their own breakpoints, which is detrimental to code quality.

The solution I chose was to export the original `width` object in `styles/device.ts` that had the original const strings for all device sizes. Then, I also made `smallerThan()` and `largerThan()` functions that were exported too. This covers all possible use-cases for breakpoints (both `min-width` and `max-width` for eight different breakpoints in `width`.)

Then I refactored all references to media queries in the client to refer to this file.

## Testing
Visual inspection shows that the website functions just like before.

## Miscellaneous
You may notice that there are still some media queries (namely in `navbar`) that are not properly implemented yet. This will be handled in the navbar PR (https://github.com/loolabs/waterpark/pull/173) in the near future.

### DIFF
--- a/client/components/Banner.tsx
+++ b/client/components/Banner.tsx
@@ -1,5 +1,5 @@
 import styled from 'styled-components'
-import { largerThan, tablet } from '../styles'
+import { largerThan, width } from '../styles'
 
 interface BannerProps {
   backgroundImageUrl: string
@@ -10,7 +10,7 @@ export const Banner = styled.div<BannerProps>`
   background-position: center;
   height: 20vh;
 
-  @media ${largerThan(tablet)} {
+  @media ${largerThan(width.tablet)} {
     height: 30vh;
   }
 `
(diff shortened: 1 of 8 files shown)


a22yuen (commented):
LGTM!

a22yuen (approved):

PR #176, opened 2021-07-13
💻 Replace mobile breakpoint

evtyz:
Assigns `width.mobile` to `425px` instead of `1024px`. Changed references from `width.mobileL` to `width.mobile` in code.

Tested visually.

### DIFF
--- a/client/styles/device.ts
+++ b/client/styles/device.ts
@@ -2,7 +2,7 @@ export const width = Object.freeze({
   mobileS: '320px',
   mobileM: '375px',
   mobileL: '425px',
-  mobile: '1024px',
+  mobile: '425px',
   tablet: '768px',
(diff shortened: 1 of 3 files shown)


client/styles/device.ts
```diff
@@ -2,7 +2,7 @@ export const width = Object.freeze({
   mobileS: '320px',
   mobileM: '375px',
   mobileL: '425px',
```
a22yuen, line 4: I think having 2 mobile sizes with the same size is redundant. We can probably remove mobileL and keep mobile since mobile is used more frequently, i.e its the standard in our codebase

a22yuen (changes requested):

a22yuen (approved):

PR #169, opened 2021-07-10
💻 Add Reviews to Resources

a22yuen:
## Screenshots
On laptops/desktops:
image

On mobile: 
image

## Purpose
Adds reviews to each resource. Closes #163 

## Approach
Changed types of Resources. This includes adding different parameters for the ratings based on the type of resource. Also add cards to display each review

Used [react-stars](https://www.npmjs.com/package/react-stars) for rating UI.

PR #177, opened 2021-07-17
💻 Increase z-index of navbar to go over gallery buttons

xujustinj:
## Screenshots

Before
Gallery uh oh

After
Gallery fixed

## Purpose

Noticed this weird visual bug while reviewing #169, here's a one-character fix.

## Testing

See screenshots


xujustinj:
Actually can we just put this in #173? @evtyz

xujustinj:
Added to #173

Issue #42, opened 2021-02-14
Static API types


jeffzh4ng:
# Purpose

It would be amazing to have shared, statically typed API types shared both by the client and server folders in our repo. Implementing this would be fully leveraging Typescript to its fullest.

# Resources
Not many resources on this. A lot of research would have to be done.

- Check out this [video](https://www.youtube.com/watch?v=GrnBXhsr0ng) by Gary Bernhardt on an end to end Typescript codebase (unfortunately it's not open source)
- https://lorefnon.tech/2018/03/25/typescript-and-validations-at-runtime-boundaries/
- io-ts might need to be used, or some sort of validator. Gary uses io-ts

xujustinj:
We will probably launch without this, then integrate it later.

PR #194, opened 2021-07-31
🗄️ Create express API type system based off OpenAPI

xujustinj:
## Purpose

Part of #42.

I kept trying to integrate the stuff from #85 with waterpark, but my changes became outdated each time we advanced `main` (or each time I decided to change the API type system).

This PR attempts to lock in a final-for-now type system. `schema.ts` defines the main structure of an `API`. Roughly speaking, it is a subset of the [OpenAPI `Paths` object](https://swagger.io/specification/#paths-object).

How does this affect you? Someday, your Waterpark types will be imported from the `/common/src/waterpark/` folder.

## Approach

The OpenAPI standard (a.k.a. Swagger, which we might use someday to externally document our APIs) is quite different from what I put in #85. The most jarring difference is that paths are _not_ recursive (`/washrooms/:id/` is not a child of `/washrooms/`). This prevents us from nesting routers (for now), but it also makes the type system more readable, and allows for flatter schema objects.

Also, OpenAPI resolves some of the discussions from #85 about whether to call something an API or an endpoint. It refers to things as "paths", where each path has one or more "operations". It sounded weird to me at first, but then it grew on me.

## Testing

Ran `npm run build` in `/common/` and it worked.

If you checkout the code, take a look at `demo.ts` to see that the compiler now gives types to the `req` and `res` objects. You can also verify that various invalid router calls get blocked by the compiler.

```ts
router.get('/washrooms/', (_, res) => {
  res.json({ this: 'is not a valid response from GET /washrooms/' })
})

router.get('/not-a-path/', (_, _) => {})

// technically still works, but req and res will contain `never` types
// because there is no POST endpoint for this path
router.post('/washrooms/:id/', (req, res) => {
  // do stuff
})
```

image

## TODOs

The code in `/common/` is not-at-all integrated with `/server/`. There are plenty of Docker nightmares coming in the future when we do integrate it, but that is not a concern for now.

Still not entirely sure what the frontend equivalent of `typexpress` would be.

### DIFF
--- a/common/src/shared/express/types.ts
+++ b/common/src/shared/express/types.ts
@@ -1,0 +1,42 @@
+import { Schema, Get } from '../api'
+import { PathParams } from 'express-serve-static-core'
+import express from 'express'
+
+type TypedHandlerForOperation<Operation extends Schema.Operation> = express.RequestHandler<
+  Get.PathParameters<Operation>,
+  Get.ResponseBody<Operation>,
+  Get.RequestBody<Operation>,
+  Get.QueryParameters<Operation>
+>
+
+export type TypedHandler<
+  API extends Schema.API,
+  PathName extends Get.PathName<API>,
+  OperationName extends Get.OperationName<API, PathName>
+> = TypedHandlerForOperation<Get.Operation<API, PathName, OperationName>>
+
+interface TypedMatcher<
+  API extends Schema.API,
+  ReturnType,
+  OperationName extends Schema.OperationName = Schema.OperationName
+> {
+  <PathName extends Get.PathName<API> & PathParams>(
+    path: PathName,
+    ...handlers: Array<
+      TypedHandler<API, PathName, OperationName & Get.OperationName<API, PathName>>
+    >
+  ): ReturnType
+}
+
+export interface TypedRouter<API extends Schema.API> {
+  all: TypedMatcher<API, this, 'all'>
+  get: TypedMatcher<API, this, 'get'>
+  post: TypedMatcher<API, this, 'post'>
+  put: TypedMatcher<API, this, 'put'>
+  delete: TypedMatcher<API, this, 'delete'>
+  patch: TypedMatcher<API, this, 'patch'>
+  options: TypedMatcher<API, this, 'options'>
+  head: TypedMatcher<API, this, 'head'>
+  // use: TypedMatcher<API, this>
+  // route: ??? (TODO)
+}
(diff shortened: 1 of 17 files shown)


common/src/waterpark/dto/washroom.ts
```diff
@@ -0,0 +1,13 @@
...
+  id: string
+  name: string
+  // TODO: the rest
+}
+
+export const tWashroom: t.Type<Washroom> = t.strict({
+  id: t.string,
+  name: t.string,
+  // TODO: the rest
+})
```
jeffzh4ng, line 13: Can you explain what `io-ts` does and why we need it again?

common/src/shared/express/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  all: TypedMatcher<API, this, 'all'>
+  get: TypedMatcher<API, this, 'get'>
+  post: TypedMatcher<API, this, 'post'>
+  put: TypedMatcher<API, this, 'put'>
+  delete: TypedMatcher<API, this, 'delete'>
+  patch: TypedMatcher<API, this, 'patch'>
+  options: TypedMatcher<API, this, 'options'>
+  head: TypedMatcher<API, this, 'head'>
+  // use: TypedMatcher<API, this>
+  // route: ??? (TODO)
```
jeffzh4ng, line 41: what's the TODO here?

common/src/shared/express/types.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface TypedRouter<API extends Schema.API> {
+  all: TypedMatcher<API, this, 'all'>
+  get: TypedMatcher<API, this, 'get'>
+  post: TypedMatcher<API, this, 'post'>
+  put: TypedMatcher<API, this, 'put'>
+  delete: TypedMatcher<API, this, 'delete'>
+  patch: TypedMatcher<API, this, 'patch'>
+  options: TypedMatcher<API, this, 'options'>
+  head: TypedMatcher<API, this, 'head'>
+  // use: TypedMatcher<API, this>
```
jeffzh4ng, line 40: why is this commented out?

jeffzh4ng (approved):

common/src/shared/express/types.ts
xujustinj: If you look at the `express.Router` type, most of the methods (`.all`, `.get`, `.post`, etc.) are covered here. This is deliberate; we're trying to have the `typexpress.Router` look, act, and quack like an Express router, so most of the Express router methods should be replicated here.

There is also `.use` which I have deliberately left out; it would be used for attaching child routers and other middleware.

`.route` is one of the remaining major methods in `Router`. I haven't figured out its type, but I also haven't figured out what it's used for. The TODO is to figure it out.

PR #198, opened 2021-07-31
💻 Add sorting to resource pages

### LINKED ISSUE #164: Sorting
Requirements:
- provide ability to sort and ~~filter~~ resources based off the following
update: we are no longer filtering, just sorting. #170

Housing:
- cleanliness
- price
- company

Study Spots:
- cleanliness
- noise

Washrooms:
- cleanliness

evtyz:
# Purpose

Closes #164.

- Adds sorting for resources based on alphabetical order, number of reviews, average reviews
- Various styling fixes
- Added in old tag code (but this is not used anywhere, we just kept it in case we want it for later)

# Screenshots
Desktop:
image

Mobile:
image


client/components/resource/Tag.tsx
```diff
@@ -1,20 +1,52 @@
...
+  line-height: ${TAG_HEIGHT};
+  margin-bottom: 1px; // make space between rows of bubbles
+  overflow: hidden;
+  padding: 0 ${TAG_HORIZONTAL_PADDING};
+  text-align: center;
+  white-space: nowrap;
+  :hover {
+    ${({ colour, highlightOnHover }) => highlightOnHover && `background-color: ${colour};`}
+  }
+`
```
jeffzh4ng, line 52: newline here too.

If we ever want to write bash scripts that process our files, it's best practice to keep our text files POSIX compliant with a newline :)

https://stackoverflow.com/a/729795/6113956

client/utils/types.ts
```diff
@@ -9,6 +7,12 @@ export const RatingCriteria: { [key: string]: Array<string> } = {
   washrooms: ['cleanliness'],
 }
 
+export const ResourceDisplayStrings: { [key: string]: string } = {
+  housing: 'Housing',
+  'study-spots': 'Study Spots',
+  washrooms: 'Washrooms',
+}
+
```
jeffzh4ng, line 15: We can also opt for one centralized data structure which is somewhat cleaner.

```
interface ResourceInfo {
  title: string
  criteria: Array<string>
}

interface ResourceLookup {
  [name: string]: ResourceInfo
}

const resourceLookup: ResourceLookup = {
  housing: {
    title: 'Housing',
    criteria: [...],
  }
}
```

But I guess the benefit with constant objects is that you can just hard-code the data right here instead of needing to create instances of the interface. What do you think @xujustinj?

client/components/resource/ResourceList.tsx
```diff
@@ -63,65 +62,153 @@ const ResourceListTitle = styled(PageTitle)`
...
+
+const SortDropdown = ({
+  slug,
+  changeSortPattern,
+}: {
+  slug: string
+  changeSortPattern: (sortPatternType) => void
+}) => {
+  let criteria = RatingCriteria[slug]
+  let sortingDefinitions = {
```
jeffzh4ng, line 112:
```suggestion
  const sortingDefinitions = {
```

client/components/resource/ResourceList.tsx
```diff
@@ -63,65 +62,153 @@ const ResourceListTitle = styled(PageTitle)`
...
+    alphabetical: (first: Resource, second: Resource) => {
+      return first.name.localeCompare(second.name) // sort from least (a) to most (z)
+    },
+    'number of ratings': (first: Resource, second: Resource) => {
+      return second.totalReviews - first.totalReviews // sort from most to least
+    },
+  }
+
+  criteria.map((item) => {
```
jeffzh4ng, line 121: I don't think map is the right function here. You use map to _map_ an array from A -> B. If you just need to loop over an array and perform some logic, use a `for...of`

client/components/resource/ResourceList.tsx
```diff
@@ -63,65 +62,153 @@ const ResourceListTitle = styled(PageTitle)`
...
+    sortingDefinitions[item] = (first: Resource, second: Resource) => {
+      return second.averageRating[item] - first.averageRating[item] // sort from most to least
+    }
+  })
+
+  criteria = criteria.map((item) => capitalizeFirstLetter(item))
+  criteria = ['Alphabetical', 'Number of Ratings', ...criteria]
```
jeffzh4ng, line 128: In general, it's best to avoid mutation like this, assigning back to the same variable and to keep things clean by creating new variables.

```
const criteria = RatingCriteria[slug]
const normalizedCriteria = criteria.map((criterion) => capitalizeFirstLetter(criterion)).unshift(['Alphabetical', 'Number of Ratings']
```

```

**Review by jeffzh4ng — CHANGES_REQUESTED** [review]
Overall great job! A few comments. Also let's see if Carolina has any feedback on the UI this weekend :)

File: client/utils/types.ts (hunk 2 above)
Comment (xujustinj) [reply client/utils/types.ts]: The interface you're suggesting here is like the API types module; a meta definition-of-a-definition. I like it. We are still exporting a const instance of the interface, and it's pretty much inline hard-coded (with type-checks that prevent you from writing incorrect inline definitions), so I don't see the benefit you mention.

If we want to go super-tryhard (mainly for our own learning's sake), we'd probably eventually hook the whole thing up with i18n. Either implementation will probably need a refactor in the future to adapt to that.

File: client/utils/types.ts [hunk 6]
```diff
@@ -1,12 +1,28 @@
...
 
+interface ResourceInfo {
+  title: string
+  criteria: Array<string>
+}
+
+interface ResourceLookup {
+  [name: string]: ResourceInfo
+}
+
```
jeffzh4ng, line 11: clean 🧼👌

jeffzh4ng (approved):

client/components/common/Functions.tsx
```diff
@@ -0,0 +1,3 @@
+export const capitalizeFirstLetter = (str: string): string => {
```
xujustinj, line 1: woah, deja vu

This change also shows up in #204, I'm guessing that's because #204 depends on this to merge first?

evtyz: Yup

PR #200, opened 2021-08-07
🗄️ Add places endpoint to initial v2 API

xujustinj:
Partial code check-in from `v2/places`. Ran unit tests locally.

This uses the New(TM) flatter module file structure. The main motivation is to make the code more accessible to people viewing it for the first time. This structure is TBD, open to suggestions,
image


KTong821 (approved):

server/src/setup/database/mikro-db.ts
```diff
@@ -48,7 +50,7 @@ class CustomNamingStrategy extends AbstractNamingStrategy implements NamingStrat
 }
 
 const clientUrl = process.env.DATABASE_URL
-// Heroku's postgres service self-signs SSL certificates,
+// Heroku's postgres service self-signs SSL certificates (whatever that means),
```
jeffzh4ng, line 53: we know what SSL certs are though!

jeffzh4ng (approved):

server/src/setup/database/mikro-db.ts
xujustinj: lol good catch, this is what happens when we check in code from a branch that is 1.5 weeks behind main

PR #205, opened 2021-08-11
💻 Fix font issues on Windows

### LINKED ISSUE #202: Fix font issues on Windows
Some fonts are not being displayed in our custom font face on Windows.

Example: ● character
image

evtyz:
# Purpose
Closes #202.

# Approach
- Explicitly import `Inter` font in `globals.css`
- Conditionally display larger ⬤ character on Windows

# Screenshots
### Before
image

### After 
image

### DIFF
--- a/client/components/resource-details/Reviews.tsx
+++ b/client/components/resource-details/Reviews.tsx
@@ -146,7 +146,8 @@ const Rating = styled(
         <RatingLabel>{capitalizeFirstLetter(label)}</RatingLabel>
         <ReactStars
           count={5}
-          char={'⬤'}
+          // Circles are too small on Windows font
+          char={navigator.appVersion.indexOf('Win') != -1 ? '⬤' : '●'} 
           value={score / 20}
           size={24}
           color1={'#DDDDDD'}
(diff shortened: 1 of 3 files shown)


client/components/resource-details/Reviews.tsx
```diff
@@ -146,7 +146,8 @@ const Rating = styled(
         <RatingLabel>{capitalizeFirstLetter(label)}</RatingLabel>
         <ReactStars
           count={5}
-          char={'●'}
```
xujustinj, line 149: I'm pretty sure there's no shortage of circle-looking Unicode characters. Have we tried using a different one? e.g. what if we use the large circle on all platforms?

xujustinj (approved):
changes lgtm but I wonder if there's a different solution altogether

client/components/resource-details/Reviews.tsx
evtyz: The large circle looks ridiculously large on Mac. Maybe too large on other platforms too.

From what I can tell, there are only two circle characters, "small black circle" and "large black circle", which can be recoloured. I might be wrong though. 

There is a "medium black circle" but it's hardcoded to be black and the colour can't be changed.

PR #210, opened 2021-08-14
💻 Upgrade next version to resolve security vulnerability

xujustinj:
Manual update to address https://github.com/loolabs/waterpark/security/dependabot/client/package.json/next/closed.

The updated Next version requires a newer version of React than we had before, so we have to upgrade that too.

We don't get automatic Dependabot fixes because we got rid of the package-lock.json files...

## Testing

The deploy preview works, which I presume means that going up a full major version of Next didn't bring about any breaking changes on our side.


evtyz (approved):
LGTM

PR #219, opened 2021-08-23
Polish v2 design (#216)

### LINKED ISSUE #216: Polish v2 frontend design
https://www.figma.com/file/AIeklk8qTC6t4rI71QtioP/Waterpark-(Wiki)-Designs?node-id=46%3A53

Address design review from @CarolinaLi
1. **Switch from full black text to a dark grey, it’ll be easier on the eyes (#222222)**
2. UW logo looks squished, also what picture will we eventually have here? Does housing have logos? Are we thinking something like this? ->
3. **Ideally everything on the page should have the same width — I think we should make the comments & top section narrower, and increase Gallery width to match that**
4. We should try to limit the number of text styles, maybe “200-300 students” and “What are your thoughts on...” could have the same style? And all buttons (add review, top nav) could also have the same styling
5. Does “200-300 students” mean the number of students living here? Where do we get this info? Was this originally from club size / should we still keep this?
6. Thoughts on removing capitalization on the year and “student”? We might want to do sentence case everywhere, like on buttons as well
7. Something like “4h ago” also feels a bit more natural than “today at 3:11 PM”
8. Would love to do rounded corners on all images, I think our icons already do this
9. For indicating the current image, having something like an underline (in brand colours) rather than an outline would make it look sleeker.
10. Nice to have: Made a different review card design to use space more efficiently. Thoughts on asking students to enter a brief comment title/summary that we can use as a card title? Would make the review card look nicer

evtyz:
Purpose: Closes #216.

### DIFF
--- a/client/components/resource-details/Gallery.tsx
+++ b/client/components/resource-details/Gallery.tsx
@@ -30,6 +30,7 @@ export const Gallery = styled(
               />
             )
           }}
+          // thumbnailPosition={'right'}
           renderRightNav={(onClick, disabled) => {
             return (
               <img
@@ -49,7 +50,4 @@ export const Gallery = styled(
   }
 )`
   width: 100%;
-  @media ${largerThan(width.laptop)} {
-    width: 60%;
-  }
 `
--- a/client/components/NavBar.tsx
+++ b/client/components/NavBar.tsx
@@ -38,7 +38,7 @@ const NavBarOption = styled.a<any>`
   ${(props: any) =>
     props.isActive &&
     css`
-      color: ${colours.black};
+      color: ${colours.darkGrey};
     `};
   margin: 0px 32px;
 
@@ -53,7 +53,7 @@ const NavBarOption = styled.a<any>`
   }
 
   &:hover {
-    color: ${colours.black};
+    color: ${colours.darkGrey};
     cursor: pointer;
   }
 `
--- a/client/components/SearchInput.tsx
+++ b/client/components/SearchInput.tsx
@@ -29,6 +29,13 @@ export const SearchInput = styled.input`
     width: 200px;
   }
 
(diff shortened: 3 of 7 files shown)


jeffzh4ng:
@evtyz marked this for ready for review by accident

CarolinaLi:
nice search bar updates! a few nits:
- can we use "sort by"?
- realized that I should've made "sort by:" and "alphabetical" different styling to help separate them. can we update the colour of "sort by:" to grey, like the search icon?
- can we also update the outline and the dropdown arrow to the same grey?
- would be nice to make "Alphabetical" bold
- on smaller screen sizes, it looks like the dropdown arrow is in the middle of the bar, can we anchor it to the right?
image

client/components/resource-details/Gallery.tsx
```diff
@@ -30,6 +30,7 @@ export const Gallery = styled(
               />
             )
           }}
+          // thumbnailPosition={'right'}
```
xujustinj, line 33:
```suggestion
```

client/components/resource-details/Reviews.tsx
```diff
@@ -94,7 +94,7 @@ const Comment = styled(({ className, review }: { className?: string; review: Rev
     <div className={className}>
       <Description>{review.comment}</Description>
       <AuthorInfo>
-        — {review.faculty} {review.status}, {formatRelative(review.timestamp, Date.now())}
+        — {review.faculty} {review.status}, {formatDistanceToNowStrict(review.timestamp)} ago
```
xujustinj, line 97: what does this do?

xujustinj (commented):
image

noticing a weird UI issue on narrow screens (~250px), not sure if we want to bother fixing it

